PR #1, opened 2022-11-05
Remove unused rerun-if causing rebuild

DSchroer:
Since `wrapper.hxx` does not exist the build script would run on every build. Adding 10 seconds per build even when the crate did not change.

### DIFF
--- a/crates/opencascade-sys/build.rs
+++ b/crates/opencascade-sys/build.rs
@@ -31,5 +31,6 @@ fn main() {
     println!("cargo:rustc-link-lib=static=wrapper");
 
     println!("cargo:rerun-if-changed=src/lib.rs");
+    println!("cargo:rerun-if-changed=include/wrapper.hxx");
     println!("cargo:rerun-if-changed=cpp/wrapper.cpp");
 }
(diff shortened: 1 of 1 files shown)


bschwind:
Oh wow, I've been meaning to look into this rebuild issue for awhile, but had powered through it by hard-coding the cmake build path in build.rs. Thanks for catching this!

crates/opencascade-sys/build.rs
```diff
@@ -31,6 +31,5 @@ fn main() {
     println!("cargo:rustc-link-lib=static=wrapper");
 
     println!("cargo:rerun-if-changed=src/lib.rs");
-    println!("cargo:rerun-if-changed=cpp/wrapper.hxx");
```
bschwind, line 34: Instead of deleting this line, we can change it to the proper path:

`println!("cargo:rerun-if-changed=include/wrapper.hxx");`

Sorry I missed this in the original 😅

DSchroer: Updated :)

bschwind (approved):
LGTM, thanks again!

PR #5, opened 2022-11-07
Expose BRepPrimAPI_MakeSphere and BRepAlgoAPI_Section.

DSchroer:
Expose BRepPrimAPI_MakeSphere and BRepAlgoAPI_Section.

### DIFF
--- a/crates/opencascade-sys/include/wrapper.hxx
+++ b/crates/opencascade-sys/include/wrapper.hxx
@@ -22,6 +22,7 @@
 #include <TopExp_Explorer.hxx>
 #include <BRepAlgoAPI_Fuse.hxx>
 #include <BRepAlgoAPI_Cut.hxx>
+#include <BRepAlgoAPI_Section.hxx>
 #include <BRepBuilderAPI_MakeEdge.hxx>
 #include <BRepBuilderAPI_MakeFace.hxx>
 #include <BRepBuilderAPI_MakeWire.hxx>
@@ -32,6 +33,7 @@
 #include <BRepMesh_IncrementalMesh.hxx>
 #include <BRepPrimAPI_MakePrism.hxx>
 #include <BRepPrimAPI_MakeBox.hxx>
+#include <BRepPrimAPI_MakeSphere.hxx>
 #include <BRepPrimAPI_MakeCylinder.hxx>
 #include <BRepOffsetAPI_MakeThickSolid.hxx>
 #include <BRepOffsetAPI_ThruSections.hxx>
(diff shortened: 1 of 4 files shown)


DSchroer:
@bschwind What are your thoughts on using `clang-format` for the C++ side so it can all be consistent?

crates/opencascade-sys/src/lib.rs
```diff
@@ -221,6 +221,10 @@ pub mod ffi {
         ) -> UniquePtr<BRepPrimAPI_MakeBox>;
         pub fn Shape(self: Pin<&mut BRepPrimAPI_MakeBox>) -> &TopoDS_Shape;
 
+        type BRepPrimAPI_MakeSphere;
+        pub fn BRepPrimAPI_MakeSphere_ctor(r: f64) -> UniquePtr<BRepPrimAPI_MakeSphere>;
+        pub fn Shape(self: Pin<&mut BRepPrimAPI_MakeSphere>) -> &TopoDS_Shape;
```
DSchroer, line 226: Since these are all inheritance based. What do you think about creating macros to cover common sub-classes? IE to expose `IsDone, Done, Shape, etc...`

bschwind:
> @bschwind What are your thoughts on using `clang-format` for the C++ side so it can all be consistent?

I've used that in the past so I think I'll set that up for this project too. Thankfully I was able to get CI working for this project, though it takes quite awhile to run.

I'll try to get C++ formatting in soon, as well as a review on this PR. Thanks!

crates/opencascade-sys/src/lib.rs
bschwind: I do think a macro is needed, but for now I've been explicitly writing everything out to observe the various patterns that fall out from the bindings.

The more cumbersome part so far has been generating constructors, and calling methods which are inherited from a base class.

I'm hoping for improvements from the `cxx` crate over time to alleviate some of these issues, but if they don't come soon then I think a macro would be a good middle ground.

crates/opencascade-sys/include/wrapper.hxx
```diff
@@ -100,13 +102,15 @@ std::unique_ptr<BRepBuilderAPI_MakeFace> BRepBuilderAPI_MakeFace_wire(const Topo
...
 std::unique_ptr<BRepPrimAPI_MakeBox> BRepPrimAPI_MakeBox_ctor(const gp_Pnt& point, double dx, double dy, double dz);
+std::unique_ptr<BRepPrimAPI_MakeSphere> BRepPrimAPI_MakeSphere_ctor(double r);
 
 // BRepLib
 bool BRepLibBuildCurves3d(const TopoDS_Shape& shape);
 
 // Boolean operations
 std::unique_ptr<BRepAlgoAPI_Fuse> BRepAlgoAPI_Fuse_ctor(const TopoDS_Shape& shape_1, const TopoDS_Shape& shape_2);
 std::unique_ptr<BRepAlgoAPI_Cut> BRepAlgoAPI_Cut_ctor(const TopoDS_Shape& shape_1, const TopoDS_Shape& shape_2);
+std::unique_ptr<BRepAlgoAPI_Section> BRepAlgoAPI_Section_ctor(const TopoDS_Shape& shape_1, const TopoDS_Shape& shape_2);
```
bschwind, line 113: Just curious, is `BRepAlgoAPI_Section` essentially a boolean intersection of two shapes, or is there something more going on with it?

bschwind: Ohhh, this page explains it pretty well:

https://dev.opencascade.org/doc/overview/html/specification__boolean_operations.html#specification__boolean_10a

crates/opencascade-sys/src/lib.rs
DSchroer: Unfortunately the cxx crate does not allow macros in the `extern "C++" ...` section. So unless we code gen it, I think its going to be manual.

bschwind (approved):
LGTM!

PR #9, opened 2022-11-12
Add BRepPrimAPI_MakeRevol

DSchroer:
Adding bindings for Add BRepPrimAPI_MakeRevol. Its useful for making complex shapes out of faces.

### DIFF
--- a/crates/opencascade-sys/cpp/wrapper.cpp
+++ b/crates/opencascade-sys/cpp/wrapper.cpp
@@ -131,6 +131,10 @@ std::unique_ptr<BRepPrimAPI_MakePrism> BRepPrimAPI_MakePrism_ctor(const TopoDS_S
   return std::unique_ptr<BRepPrimAPI_MakePrism>(new BRepPrimAPI_MakePrism(shape, vec, copy, canonize));
 }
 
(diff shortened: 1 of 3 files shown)


crates/opencascade-sys/cpp/wrapper.cpp
```diff
@@ -131,6 +131,10 @@ std::unique_ptr<BRepPrimAPI_MakePrism> BRepPrimAPI_MakePrism_ctor(const TopoDS_S
   return std::unique_ptr<BRepPrimAPI_MakePrism>(new BRepPrimAPI_MakePrism(shape, vec, copy, canonize));
 }
 
+std::unique_ptr<BRepPrimAPI_MakeRevol> BRepPrimAPI_MakeRevol_ctor(const TopoDS_Shape& shape, const gp_Ax1& axis, const Standard_Real angle, const Standard_Boolean copy){
```
bschwind, line 134: I missed this on the previous review, but feel free to just use `double` instead of `Standard_Real`, and `bool` instead of `Standard_Boolean`.

DSchroer: I kind of prefer to use the ones they use in the docs. Lowers the cognitive complexity since you can copy paste signatures.

bschwind (approved):
LGTM, thank you!

PR #10, opened 2022-11-15
Add BRep_Tool{Pnt, Curve} and HandleGeomCurve_Value

DSchroer:
Add BRep_Tool{Pnt, Curve} and HandleGeomCurve_Value.
Tools for accessing value within vertexes and edges.

### DIFF
--- a/crates/opencascade-sys/build.rs
+++ b/crates/opencascade-sys/build.rs
@@ -14,7 +14,8 @@ fn main() {
         .define("USE_FFMPEG", "FALSE")
         .build();
 
-    println!("cargo:rustc-link-search=native={}", dst.join("lib").display());
+    println!("cargo:rustc-link-search=native={}", dst.join("win64/gcc/lib").display()); // needed for windows builds
+    println!("cargo:rustc-link-search=native={}", dst.join("lib").display()); // needed for windows builds
     println!("cargo:rustc-link-lib=static=TKMath");
     println!("cargo:rustc-link-lib=static=TKernel");
     println!("cargo:rustc-link-lib=static=TKGeomBase");
@@ -35,7 +36,8 @@ fn main() {
 
     cxx_build::bridge("src/lib.rs")
         .cpp(true)
-        .flag_if_supported("-std=c++11")
+        .flag_if_supported("-std=c++14")
+        .include(format!("{}", dst.join("inc").display())) // needed for windows builds
         .include(format!("{}", dst.join("include").join("opencascade").display()))
         .include("include")
         .file("cpp/wrapper.cpp")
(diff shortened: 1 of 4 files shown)


crates/opencascade-sys/build.rs
```diff
@@ -35,7 +35,7 @@ fn main() {
 
     cxx_build::bridge("src/lib.rs")
         .cpp(true)
-        .flag_if_supported("-std=c++11")
+        .flag_if_supported("-std=c++14")
```
bschwind, line 38: I'd prefer if we don't change this just for the sake of being able to use `make_unique`.

DSchroer: Is there any reason to use c++11? Its still a 11 year old standard at this point.

bschwind: My reasoning is just that if it works with c++11 already, there's no reason to raise the requirements arbitrarily. I don't imagine we'll be writing a lot of exotic C++ code, it's mostly just wrappers for constructors.

C++11 has been around long enough that it's well supported by all the major compilers out there. C++14 is probably around that point as well, but I'd rather not move to it unless there's a very compelling reason to do so.

DSchroer: Ill push an update that works for C++11. Thanks for explaining the reasoning.

bschwind: Sorry for the churn on that.

crates/opencascade-sys/cpp/wrapper.cpp
```diff
@@ -10,6 +10,10 @@ rust::String type_name(const HandleStandardType& handle) {
   return std::string(handle->Name());
 }
 
+std::unique_ptr<gp_Pnt> HandleGeomCurve_Value(const HandleGeomCurve& curve, const Standard_Real U) {
+  return make_unique<gp_Pnt>(curve->Value(U));
```
bschwind, line 14: Sorry to beat a dead horse, but is there a particular reason you need `make_unique` here vs. just using `return std::unique_ptr<T>(...);` as it exists elsewhere in the code?

I don't code in C++ in my day-to-day so maybe I'm missing something.

bschwind:
```suggestion
  return std::unique_ptr<gp_Pnt>(new gp_Pnt(curve->Value(U)));
```

We should be able to use it in this way without needing to create the `make_unique` util function. Sorry if it seems old fashioned, but I'd just prefer all the wrapper code to follow the same pattern so we can change them all at once later with less confusion on why one function used `make_unique` and another used the `unique_ptr` constructor directly.

DSchroer: Updated

bschwind (approved):
Thanks again and sorry for the churn!

PR #13, opened 2022-12-21
Add IsDone to all Tools and IsNull to all Handles

DSchroer:
Add IsDone to all Tools and IsNull to all Handles.

Tools that do not check for IsDone will throw.
Handles that do not check IsNull may segfault. 
These are needed to wrap those operations safely.


DSchroer:
@bschwind do you mind re-running the cargo clippy job? It failed on repo checkout.

bschwind:
@DSchroer Re-running now, and I'll get to reviewing these later.

bschwind (approved):
LGTM, thank you!

PR #14, opened 2022-12-22
Add support to retrieve 3D triangle data without writing to STL

DSchroer:
Adding the necessary parts to access mesh 3D triangle data. Supports triangles, nodes and normals. 

I also added a test case to the project. This was mainly to ensure that what I was adding was going to work. I think it cant hurt to leave in but feel free to let me know if you prefer I didn't include it. 

Note: I tried a different way of dealing with handles. Using the `get` property and making an unsafe case to a pointer. This should be an easier pattern since it doesn't need so much custom implementation within the wrapper.

### DIFF
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -458,7 +458,7 @@ pub mod ffi {
         /// # Safety
         /// Since this handle may be null, this is only safe if IsNull has been checked.
         /// The handle must also outlive the reference returned.
-        pub unsafe fn get(self: &Handle_Poly_Triangulation) -> *mut Poly_Triangulation;
+        pub fn get(self: &Handle_Poly_Triangulation) -> *mut Poly_Triangulation;
 
         type Poly_Triangulation;
         pub fn NbNodes(self: &Poly_Triangulation) -> i32;
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -455,9 +455,6 @@ pub mod ffi {
 
         type Handle_Poly_Triangulation;
         pub fn IsNull(self: &Handle_Poly_Triangulation) -> bool;
-        /// # Safety
-        /// Since this handle may be null, this is only safe if IsNull has been checked.
-        /// The handle must also outlive the reference returned.
         pub fn get(self: &Handle_Poly_Triangulation) -> *mut Poly_Triangulation;
 
         type Poly_Triangulation;
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -556,11 +556,15 @@ pub mod ffi {
         pub fn IsDone(self: &BRepMesh_IncrementalMesh) -> bool;
 
         type TopLoc_Location;
+        #[cxx_name = "construct_unique"]
         pub fn TopLoc_Location_ctor() -> UniquePtr<TopLoc_Location>;
 
         type Handle_Poly_Triangulation;
         pub fn IsNull(self: &Handle_Poly_Triangulation) -> bool;
-        pub fn get(self: &Handle_Poly_Triangulation) -> *mut Poly_Triangulation;
+        #[cxx_name = "handle_try_deref"]
+        pub fn Handle_Poly_Triangulation_Get(
+            handle: &Handle_Poly_Triangulation,
+        ) -> Result<&Poly_Triangulation>;
 
         type Poly_Triangulation;
         pub fn NbNodes(self: &Poly_Triangulation) -> i32;
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -13,6 +13,14 @@ pub mod ffi {
         TopAbs_SHAPE,
     }
 
+    #[repr(u32)]
+    pub enum TopAbs_Orientation {
+        TopAbs_FORWARD,
+        TopAbs_REVERSED,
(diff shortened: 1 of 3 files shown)


crates/opencascade-sys/tests/triangulation.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let mut edge_explorer =
+        TopExp_Explorer_ctor(mesh.pin_mut().Shape(), TopAbs_ShapeEnum::TopAbs_FACE);
+    while edge_explorer.More() {
+        let face = TopoDS_cast_to_face(edge_explorer.Current());
+        let mut location = TopLoc_Location_ctor();
+
+        let triangulation_handle = BRep_Tool_Triangulation(&face, location.pin_mut());
+        if !triangulation_handle.IsNull() {
+            let triangulation = unsafe { &*triangulation_handle.get() };
```
bschwind, line 24: Hmmm, yeah I see what you mean here. I would really like to avoid introducing more `unsafe` functions than is necessary, preferably keeping that layer of code just to `cxx`.

Having particular requirements on calling the function can make it error prone (null check before use, make sure the handle outlives the returned reference, etc.). We should provide an API that doesn't have these requirements, either by writing the more verbose code that uses `cxx`, or providing some abstraction that uses type safety and lifetimes to guarantee it can't be misused.

DSchroer: I feel that its pretty normal to have unsafe in a `-sys` crate. Since we are both exposing safe wrappers anyway, I feel its better to do this than to write more glue in C++. This makes cleaner and way more scalable.

bschwind: True, `-sys` crates typically are where the `unsafe` code lives. Currently I'm relying on `cxx` to hide that from the bindings.

I haven't gone through the process of trying to code what you're doing in this PR, so I'll try to go through that and understand the needs here - we might be able to come up with something that is both safe, and convenient to generate and use.

DSchroer: So small update. I was thinking about safety here. Really the get() method itself isnt unsafe and we I removed the unsafe marker from cxx. What is unsafe is casting a pointer to a reference without a null check. So using the result of `get` still needs to be in an unsafe block. How about we expose this for now since it reduces glue code and we can hide it if we decided to provide better type wrappers in the `-sys` create.

DSchroer: Have you had a chance to look at this?

bschwind: @DSchroer sorry I haven't had time to look closely at this and try to find alternatives. I'll try to find a bit of time today to experiment with this branch but no guarantees.

bschwind: @DSchroer What I'll try to do for now is reduce the overall burden of the C++ wrapper code by implementing your suggestions in the issues you've filed.

I know I was against introducing C++ templates in the past, but maybe there's a clever one we can come up with to deal with opencascade `Handle`s in an ergonomic way, where we can still have the safety guarantees from `cxx` without having to write lots of wrapper code.

DSchroer: Ive added a template that allows us to safely bind handles to `xxx -> Result<&T>` functions. Please let me know what you think.

bschwind: @DSchroer looks great! Also really cool that CXX does the translation to a rust function which returns a `Result`, nice.

crates/opencascade-sys/src/lib.rs
```diff
@@ -179,6 +187,10 @@ pub mod ffi {
         pub fn IsNull(self: &TopoDS_Shape) -> bool;
         pub fn IsEqual(self: &TopoDS_Shape, other: &TopoDS_Shape) -> bool;
 
+        type TopAbs_Orientation;
+        pub fn Orientation(self: &TopoDS_Shape) -> TopAbs_Orientation;
+        pub fn Orientation(self: &TopoDS_Face) -> TopAbs_Orientation;
```
bschwind, line 192: Are these currently used or needed?

crates/opencascade-sys/src/lib.rs
```diff
@@ -534,5 +554,33 @@ pub mod ffi {
 
         pub fn Shape(self: &BRepMesh_IncrementalMesh) -> &TopoDS_Shape;
         pub fn IsDone(self: &BRepMesh_IncrementalMesh) -> bool;
+
+        type TopLoc_Location;
+        #[cxx_name = "construct_unique"]
+        pub fn TopLoc_Location_ctor() -> UniquePtr<TopLoc_Location>;
+
+        type Handle_Poly_Triangulation;
+        pub fn IsNull(self: &Handle_Poly_Triangulation) -> bool;
```
bschwind, line 563: This can probably be removed unless you think we'll be using it.

bschwind (approved):
This is looking great, thanks for the iterations on it!

I left a few comments about potentially unneeded code, but those are minor nitpicks.

crates/opencascade-sys/src/lib.rs
DSchroer: They are needed. Its not used in the test but if you want to calculate the normals for each triangle you need this info,

crates/opencascade-sys/src/lib.rs
DSchroer: I feel like leaving it is better. Technically there could be many reasons that the request fails but using IsNull is clear enough.

bschwind:
Thanks again, @DSchroer!

PR #16, opened 2022-12-24
Unify the explorer interface using XXX_to_owned

DSchroer:
Currently the cast methods return either `&Something` or `UniquePtr<Something>` depending on convenience. This makes it hard to make a generic wrapper around the Explorer class since those values have different lifetimes. 

Here I make all casts return references and add a suite of `Something_to_owned` methods that can make those values `UniquePtr<Something>` on request. 

I also remove `new_shape` since it is now obsolete.


bschwind (approved):
Looks great! Just needs a rebase on the new header-only commits in `main`.

PR #18, opened 2022-12-27
Move to a header only workflow to reduce boilerplate

DSchroer:
Rather than triple define each C++ function, move to a header only system that gives us just as much control for the code we are writing. 

Also added .clang-format and Clang Format workflows to keep the C++ code as consistent as the rust code.


bschwind:
@DSchroer sorry, stepped away from most of my projects for a few weeks for the holidays. I'm back now and will try to get back to these reviews this weekend.

DSchroer:
@bschwind Thanks. Let me know when you have a change. Ive been avoiding any more changes until we get through the backlog.

bschwind (approved):
This looks great! I always disliked typing (copy-pasting) the same function definition twice.

I might have a followup PR for some C++ formatting tweaks (mostly changing it to 4 spaces for indentation) but it's very minor.

Issue #17, opened 2022-12-26
Proposal: Consistent naming / patterns


DSchroer:
Currently the naming of methods and functions in lib.rs is inconsistent:
- `type_name`
- `HandleGeomCurve_Value`
- `DynamicType`

Proposal:
## Naming
1. Always use the Opencascade convention for C++. (`CammelCase` with no `_`, including standalone functions). Not the rust convention.
```cxx
type_name(...) -> TypeName(...)
```

2. Since `cxx` has no better way yet, use `_` to show associated types for functions that cant be bound using `self`. This covers constructors and other similar functions.
```
HandleGeomCurve_Value
```

3. In `lib.rs`, since it is in rust, we always use rust naming conventions and make use of the `cxx_name` attribute to point to the correct `c++` method.
```rs
#[cxx_name = "Shape"]
pub fn shape(self: Pin<&mut BRepAlgoAPI_Fuse>) -> &TopoDS_Shape;
```

## Access
Free floating functions are hard to reason about. Since Opencascade is a class based library, we should follow the same by re-attaching methods to the respective parent. If a function is associated with a type it should be private and attached via an impl block to the associated type. 

```rs
...
    type StlAPI_Writer;
    fn StlAPI_Writer_ctor() -> UniquePtr<StlAPI_Writer>;
  }
}
...

impl ffi::StlAPI_Writer {
    pub fn new() -> cxx::UniquePtr<ffi::StlAPI_Writer> {
        ffi::StlAPI_Writer_ctor()
    }
}
```

What do you think? I feel this will be a good step forward and help reason about how future additions to the `-sys` crate should be made.

bschwind:
This sounds reasonable to me!

I hadn't put much thought into organization as you can probably tell, my original goal was to just get this working on the bottle tutorial and get a feel for how to bind to C++ code.

You're right to be thinking about better organization now though so it doesn't become a pain.

Maybe the best way to start is to apply what you've suggested to one of the opencascade types and open a PR so we can get a feel for how it'll look.

Eventually I want to make a much higher level layer on top of all this, but it will be very helpful to more easily bind to the lower level stuff with your suggestions.

PR #19, opened 2023-02-11
Constructor template

### LINKED ISSUE #15: Proposal: Use a contructor template
I know in the past you have opted against using templates on the C++ side. However I was browsing the `cxx` thread regarding constructors and one of the recent posts recommended a workaround to the problem. Rather than instrument each constructor by itself, create a template like so:
```
template<typename T, typename... Args>
std::unique_ptr<T> construct(Args... args)
{
  return std::unique_ptr(new T(args...));
}
```

Then we can just re-use it when binding a new constructor.

What are your thoughts on taking this approach to remove some of the boilerplate?

Original thread:
https://github.com/dtolnay/cxx/issues/280#issuecomment-1344153115

bschwind:
Closes #15

For now I kept the function names the same, but when we address #17 we can clean up the naming conventions.

@DSchroer what do you think about also using this template for the `*_to_owned()` functions?

### DIFF
--- a/crates/opencascade-sys/include/wrapper.hxx
+++ b/crates/opencascade-sys/include/wrapper.hxx
@@ -44,7 +44,6 @@
 
 // Generic template constructor
 template <typename T, typename... Args> std::unique_ptr<T> construct_unique(Args... args) {
-  // return T(args...);
   return std::unique_ptr<T>(new T(args...));
 }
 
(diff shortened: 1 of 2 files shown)


crates/opencascade-sys/include/wrapper.hxx
```diff
@@ -42,6 +42,12 @@
 #include <gp_Trsf.hxx>
 #include <gp_Vec.hxx>
 
+// Generic template constructor
+template <typename T, typename... Args> std::unique_ptr<T> construct_unique(Args... args) {
+  // return T(args...);
```
DSchroer, line 47: Is this comment needed?

crates/opencascade-sys/src/lib.rs
```diff
@@ -21,7 +21,9 @@ pub mod ffi {
 
         // Runtime
         type Message_ProgressRange;
-        pub fn Message_ProgressRange_ctor() -> UniquePtr<Message_ProgressRange>;
+
+        #[rust_name = "Message_ProgressRange_ctor"]
+        pub fn construct_unique() -> UniquePtr<Message_ProgressRange>;
```
DSchroer, line 26: Why not go the other way and use `cxx_name` attributes. Since you are writing rust its better to keep the rust side 1:1 no?

DSchroer (changes requested):
Looks great as an implementation. Just some minor suggestions.

DSchroer:
For `_to_owned` I think this works well too. Good plan.

crates/opencascade-sys/include/wrapper.hxx
bschwind: Nope that's just me forgetting to remove it :)

crates/opencascade-sys/src/lib.rs
bschwind: Oh, good call! I'll make this change.

DSchroer (approved):
LGTM!

PR #20, opened 2023-02-24
Expose the GProp API

DSchroer:
Expose the GProp API so that we can compute things like the center of mass of an object.


bschwind (approved):
LGTM!

PR #26, opened 2023-05-28
Add dynamic linking support for faster builds

DSchroer:
Working out of multiple repositories, the `opencascade-sys` library tends to be a major bottleneck for building. Added a `dynamic` feature that uses system libs rather than building opencascade from scratch each time.


bschwind (commented):
Thanks for the PR! There's a bit more work to be done to make this portable but it's going in the right direction.

bschwind:
Closed in favor of #87

PR #74, opened 2023-07-10
Add keycap example from OPK

bschwind:
Closes #23

PR #76, opened 2023-07-11
Chamfer support to primitives

### LINKED ISSUE #39: Support chamfers
Reference: https://dev.opencascade.org/doc/overview/html/occt_user_guides__modeling_algos.html#occt_modalg_6_1_2

### LINKED ISSUE #78: Chamfer the edges of the keycap stem cross
Related to #23

This is the last step (aside from text legends) to replicate the OPK keycap generator.

What we have:

Screen Shot 2023-07-12 at 15 35 29

What we want:
Screen Shot 2023-07-12 at 15 35 42

tuna-f1sh:
I wanted to get up to speed with the Open CASCADE stuff so thought I'd give https://github.com/bschwind/opencascade-rs/issues/39 a go. I saw that the `chamfer_edges` is in the lib `Shape` but this adds to the primitives. It uses the `AddChamfer` function from the `MakeFillet2d`.

There are a couple of bits I'm not sure of so I'll comment at those lines. Tested by adding

```
    let mut shell_bottom = Workplane::xy().rect(bx - thickness * 2.0, by - thickness * 2.0);
    shell_bottom.chamfer_angle(2.0, 0.78);

    let mut shell_mid = Workplane::xy()
        .transformed(dvec3(0.0, 0.0, height / 4.0), dvec3(0.0, 0.0, 0.0))
        .rect(bx - thickness * 3.0, by - thickness * 3.0);
    shell_mid.chamfer_angle(1.0, 0.78);

    let mut shell_top = Workplane::xy()
        .transformed(dvec3(0.0, 0.0, height - height / 4.0 - 4.5), dvec3(angle, 0.0, 0.0))
        .rect(tx - thickness * 2.0 + 0.5, ty - thickness * 2.0 + 0.5);
    shell_top.chamfer_angle(1.0, 0.78);
```

to the keycap one can get a nice tapering chamfer. I think you said the keycap cross requires a chamfer too? I didn't want to change that too much whilst https://github.com/bschwind/opencascade-rs/pull/74 is open.

Screenshot 2023-07-11 at 17 28 22

Resolves #39
Resolves #78

### DIFF
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -466,6 +466,20 @@ pub mod ffi {
             vertex: &TopoDS_Vertex,
             radius: f64,
         ) -> UniquePtr<TopoDS_Edge>;
+        pub fn BRepFilletAPI_MakeFillet2d_add_chamfer(
+            make_fillet: Pin<&mut BRepFilletAPI_MakeFillet2d>,
+            edge1: &TopoDS_Edge,
+            edge2: &TopoDS_Edge,
+            distance1: f64,
+            distance2: f64,
+        ) -> UniquePtr<TopoDS_Edge>;
+        pub fn BRepFilletAPI_MakeFillet2d_add_chamfer_angle(
+            make_fillet: Pin<&mut BRepFilletAPI_MakeFillet2d>,
+            edge: &TopoDS_Edge,
+            vertex: &TopoDS_Vertex,
+            distance: f64,
+            angle: f64,
+        ) -> UniquePtr<TopoDS_Edge>;
         pub fn Build(self: Pin<&mut BRepFilletAPI_MakeFillet2d>, progress: &Message_ProgressRange);
         pub fn Shape(self: Pin<&mut BRepFilletAPI_MakeFillet2d>) -> &TopoDS_Shape;
         pub fn IsDone(self: &BRepFilletAPI_MakeFillet2d) -> bool;
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -134,6 +134,12 @@ pub mod ffi {
             child_type: TopAbs_ShapeEnum,
             shape_data_map: Pin<&mut TopTools_IndexedDataMapOfShapeListOfShape>,
         );
+        pub fn map_shapes_and_unique_ancestors(
+            shape: &TopoDS_Shape,
+            parent_type: TopAbs_ShapeEnum,
+            child_type: TopAbs_ShapeEnum,
+            shape_data_map: Pin<&mut TopTools_IndexedDataMapOfShapeListOfShape>,
+        );
 
         type TColgp_Array1OfDir;
         #[cxx_name = "construct_unique"]
@@ -749,6 +755,24 @@ pub mod ffi {
         pub fn Next(self: Pin<&mut TopExp_Explorer>);
         pub fn ExplorerCurrentShape(explorer: &TopExp_Explorer) -> UniquePtr<TopoDS_Shape>;
         pub fn Current(self: &TopExp_Explorer) -> &TopoDS_Shape;
+
+        pub fn TopExp_FirstVertex(edge: &TopoDS_Edge) -> UniquePtr<TopoDS_Vertex>;
+        pub fn TopExp_LastVertex(edge: &TopoDS_Edge) -> UniquePtr<TopoDS_Vertex>;
+        pub fn TopExp_EdgeVertices(
+            edge: &TopoDS_Edge,
+            vertex_first: Pin<&mut TopoDS_Vertex>,
+            vertex_last: Pin<&mut TopoDS_Vertex>,
+        );
+        pub fn TopExp_WireVertices(
+            wire: &TopoDS_Wire,
+            vertex_first: Pin<&mut TopoDS_Vertex>,
+            vertex_last: Pin<&mut TopoDS_Vertex>,
+        );
+        pub fn TopExp_CommonVertex(
+            edge_1: &TopoDS_Edge,
+            edge_2: &TopoDS_Edge,
+            vertex: Pin<&mut TopoDS_Vertex>,
+        ) -> bool;
 
         pub fn BRep_Tool_Surface(face: &TopoDS_Face) -> UniquePtr<HandleGeomSurface>;
         pub fn BRep_Tool_Curve(
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -32,6 +32,14 @@ pub mod ffi {
         IFSelect_RetStop,
     }
 
+    #[derive(Debug)]
+    #[repr(u32)]
+    pub enum BOPAlgo_GlueEnum {
+        BOPAlgo_GlueOff,
+        BOPAlgo_GlueShift,
+        BOPAlgo_GlueFull,
+    }
+
     unsafe extern "C++" {
         // https://github.com/dtolnay/cxx/issues/280
 
@@ -616,6 +624,7 @@ pub mod ffi {
 
         // Boolean Operations
         type BRepAlgoAPI_Fuse;
+        type BOPAlgo_GlueEnum;
 
         #[cxx_name = "construct_unique"]
         pub fn BRepAlgoAPI_Fuse_ctor(
@@ -627,6 +636,7 @@ pub mod ffi {
         pub fn Build(self: Pin<&mut BRepAlgoAPI_Fuse>, progress: &Message_ProgressRange);
         pub fn IsDone(self: &BRepAlgoAPI_Fuse) -> bool;
         pub fn SectionEdges<'a>(self: Pin<&'a mut BRepAlgoAPI_Fuse>) -> &'a TopTools_ListOfShape;
+        pub fn SetGlue(self: Pin<&mut BRepAlgoAPI_Fuse>, glue: BOPAlgo_GlueEnum);
 
         type BRepAlgoAPI_Cut;
 
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -123,17 +123,17 @@ pub mod ffi {
             self: &TopTools_IndexedDataMapOfShapeListOfShape,
             index: i32,
         ) -> &TopoDS_Shape;
-        pub fn FindFromIndex<'a>(
-            self: &'a TopTools_IndexedDataMapOfShapeListOfShape,
+        pub fn FindFromIndex(
+            self: &TopTools_IndexedDataMapOfShapeListOfShape,
             index: i32,
-        ) -> &'a TopTools_ListOfShape;
+        ) -> &TopTools_ListOfShape;
         pub fn FindIndex(
             self: &TopTools_IndexedDataMapOfShapeListOfShape,
             shape: &TopoDS_Shape,
         ) -> i32;
         pub fn FindFromKey<'a>(
             self: &'a TopTools_IndexedDataMapOfShapeListOfShape,
-            shape: &TopoDS_Shape,
+            shape: &'a TopoDS_Shape,
         ) -> &'a TopTools_ListOfShape;
 
         pub fn map_shapes_and_ancestors(
@@ -635,7 +635,7 @@ pub mod ffi {
         pub fn Shape(self: Pin<&mut BRepAlgoAPI_Fuse>) -> &TopoDS_Shape;
         pub fn Build(self: Pin<&mut BRepAlgoAPI_Fuse>, progress: &Message_ProgressRange);
         pub fn IsDone(self: &BRepAlgoAPI_Fuse) -> bool;
-        pub fn SectionEdges<'a>(self: Pin<&'a mut BRepAlgoAPI_Fuse>) -> &'a TopTools_ListOfShape;
+        pub fn SectionEdges(self: Pin<&mut BRepAlgoAPI_Fuse>) -> &TopTools_ListOfShape;
         pub fn SetGlue(self: Pin<&mut BRepAlgoAPI_Fuse>, glue: BOPAlgo_GlueEnum);
 
         type BRepAlgoAPI_Cut;
@@ -653,7 +653,7 @@ pub mod ffi {
             self: Pin<&'a mut BRepAlgoAPI_Cut>,
             shape: &'a TopoDS_Shape,
         ) -> &'a TopTools_ListOfShape;
-        pub fn SectionEdges<'a>(self: Pin<&'a mut BRepAlgoAPI_Cut>) -> &'a TopTools_ListOfShape;
+        pub fn SectionEdges(self: Pin<&mut BRepAlgoAPI_Cut>) -> &TopTools_ListOfShape;
 
         type BRepAlgoAPI_Common;
 
--- a/crates/opencascade-sys/src/lib.rs
+++ b/crates/opencascade-sys/src/lib.rs
@@ -105,6 +105,36 @@ pub mod ffi {
             shape_map: Pin<&mut TopTools_IndexedMapOfShape>,
         );
 
+        type TopTools_IndexedDataMapOfShapeListOfShape;
+
+        #[cxx_name = "construct_unique"]
+        pub fn new_indexed_data_map_of_shape_list_of_shape(
+        ) -> UniquePtr<TopTools_IndexedDataMapOfShapeListOfShape>;
+        pub fn Extent(self: &TopTools_IndexedDataMapOfShapeListOfShape) -> i32;
+        pub fn FindKey(
+            self: &TopTools_IndexedDataMapOfShapeListOfShape,
+            index: i32,
+        ) -> &TopoDS_Shape;
+        pub fn FindFromIndex<'a>(
+            self: &'a TopTools_IndexedDataMapOfShapeListOfShape,
+            index: i32,
+        ) -> &'a TopTools_ListOfShape;
+        pub fn FindIndex(
+            self: &TopTools_IndexedDataMapOfShapeListOfShape,
+            shape: &TopoDS_Shape,
+        ) -> i32;
+        pub fn FindFromKey<'a>(
+            self: &'a TopTools_IndexedDataMapOfShapeListOfShape,
+            shape: &TopoDS_Shape,
+        ) -> &'a TopTools_ListOfShape;
+
+        pub fn map_shapes_and_ancestors(
+            shape: &TopoDS_Shape,
+            parent_type: TopAbs_ShapeEnum,
+            child_type: TopAbs_ShapeEnum,
+            shape_data_map: Pin<&mut TopTools_IndexedDataMapOfShapeListOfShape>,
+        );
+
         type TColgp_Array1OfDir;
         #[cxx_name = "construct_unique"]
         pub fn TColgp_Array1OfDir_ctor(
@@ -590,6 +620,7 @@ pub mod ffi {
         pub fn Shape(self: Pin<&mut BRepAlgoAPI_Fuse>) -> &TopoDS_Shape;
         pub fn Build(self: Pin<&mut BRepAlgoAPI_Fuse>, progress: &Message_ProgressRange);
         pub fn IsDone(self: &BRepAlgoAPI_Fuse) -> bool;
+        pub fn SectionEdges<'a>(self: Pin<&'a mut BRepAlgoAPI_Fuse>) -> &'a TopTools_ListOfShape;
 
         type BRepAlgoAPI_Cut;
 
--- a/crates/opencascade/examples/chamfer.rs
+++ b/crates/opencascade/examples/chamfer.rs
@@ -1,6 +1,6 @@
 use glam::dvec3;
 use opencascade::{
-    primitives::{Direction, Edge, Face, Solid},
+    primitives::{Direction, Face, Solid},
     workplane::Workplane,
 };
 
@@ -15,11 +15,13 @@ pub fn main() {
 
     let chamfered_box = Solid::loft([&base, &top].into_iter());
 
-    let handle = Workplane::xy().rect(5.0, 5.0);
+    // insert the workplane into the chamfered box area so union returns edges
+    let handle = Workplane::xy().translated(dvec3(0.0, 0.0, 0.1)).rect(5.0, 5.0);
     let handle_face = Face::from_wire(&handle);
 
     let handle_body = handle_face.extrude(dvec3(0.0, 0.0, -10.0));
-    let (mut chamfered_shape, _fuse_edges) = chamfered_box.union(&handle_body);
+    let (mut chamfered_shape, fuse_edges) = chamfered_box.union(&handle_body);
+    chamfered_shape.chamfer_edges(0.5, &fuse_edges);
 
     // chamfer the top of the protrusion
     let top_edges: Vec<_> = chamfered_shape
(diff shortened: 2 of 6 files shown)


strohel (commented):
Wow, you were quick to get around the codebase and submit something substantial!

Are the implementations of `chamger{,_angle,edge,edges}()` inspired by some existing code, or is that from scratch and unique to opencascase-rs? I'm not sure if I'd be able to come up with them with my current understanding.

Added one supernit for the sake of it, but almost ashamed how trivial it is. :upside_down_face:

strohel:
(edited description to automatically close #39 when merged, hope that's correct)

tuna-f1sh:
> Are the implementations of chamger{,_angle,edge,edges}() inspired by some existing code, or is that from scratch and unique to opencascase-rs? I'm not sure if I'd be able to come up with them with my current understanding.

They are mostly inspired by the existing code. `edge`, `edges` are almost 1:1 the fillet functions but using the [MakeChamfer](https://dev.opencascade.org/doc/overview/html/occt_user_guides__modeling_algos.html#occt_modalg_6_1_2) - it said in the docs it was similar so thought it was a good starting point! The angle was a bit different and required getting my head around the Wire, Faces, Edges, Shape but the tutorial was useful for that. Still not 100% so it's more of a review request than perfect PR!

 It is quite a hard code base to jump into so thought some feedback early would be helpful.

bschwind:
Chamfers are overall a bit trickier than fillets, I'm realizing. I moved part of your example code to the viewer app to more easily view the faces that get generated. Here's the code:

```rust
let mut base = Workplane::xy().rect(10.0, 10.0);
base.chamfer_angle(2.0, 3.14159 / 8.0);
```

Screen Shot 2023-07-12 at 13 28 14

It's a bit asymmetric and I'm not sure if that's expected or not.

This works a bit better:

```rust
let mut base = Workplane::xy().rect(10.0, 10.0);
base.chamfer(1.0, Some(2.0));
```

Screen Shot 2023-07-12 at 13 31 56

I get the feeling if we want to support non 45-degree chamfers we'll need to provide more control in the API to determine which angles are associated with which edges.

I would be fine for now to simplify this PR to just include simple 45 degree chamfers which _should_ work as expected in all scenarios.

Not sure if you already know, but since we're quite close to cadquery in terms of API at the moment, there is no shame in looking at their implementation as well (everything these projects are built on are all open source)

https://github.com/CadQuery/cadquery/blob/76dbeef63007d40e7a2003a444552ceeeb1367ad/cadquery/occ_impl/shapes.py#L2536-L2557

bschwind (commented):
This is looking great so far! I feel bad to make you wade through the currently very messy code, but I'm hoping to get it into a slightly better state soon.

Regarding the actual binding code, I think you did it perfectly for how the codebase is currently set up. It's always unfortunate when we have to write more C++ code in `wrapper.hxx`, so I'm hoping we can eventually figure out a way to use template magic to call class methods and return results in `UniquePtr`, similar to what we do with `construct_unique`.

As I mentioned, I would be okay with tweaking this PR to only include 45-degree chamfers for now, as that simplifies things a good amount and we can then work more on how the more complicated chamfering operations would look from an API point of view.

tuna-f1sh:
@bschwind I've merged main back into this and updated the `Edge` impl so it has the fillet and chamfer methods. I also added that `Angle` suggestion from you but only for the chamfer. I think it needs a separate PR to adapt all functions that take a degree/radian - https://github.com/bschwind/opencascade-rs/blob/chamfer/crates/opencascade/src/primitives.rs#L338 is a good candidate amongst others.

As you say, the `chamfer_angle` on all edges it is little weird and probably not very useful. I think the use case for that function would be for a specific edge instead. So perhaps I should move it to the `Edge` impl.

bschwind:
@tuna-f1sh thanks!

> I also added that Angle suggestion from you but only for the chamfer operations. I think it needs a separate PR to adapt all functions that take a degree/radian

Thanks for starting it, I agree it makes sense to be a separate PR. I'd be happy to start that!

> I think the use case for that function would be for a specific edge instead. So perhaps I should move it to the Edge impl.

I _think_ it would still need to be a method on `Shape` or `Solid`, because I believe Edges can be shared among different shapes. I'm not 100% sure on that, we'd probably need to do some experimentation.

crates/opencascade/examples/chamfer.rs
```diff
@@ -0,0 +1,47 @@
...
+    let handle_face = Face::from_wire(&handle);
+
+    let handle_body = handle_face.extrude(dvec3(0.0, 0.0, -10.0));
+    let mut chamfered_shape = chamfered_box.union(&handle_body);
+
+    // chamfer the top of the protrusion by getting shape edges at z = -10.0
+    let top_edges = chamfered_shape
+        .edges()
+        .filter(|edge| edge.start_point().z == -10.0 && edge.end_point().z == -10.0)
+        .collect::<Vec<Edge>>();
```
bschwind, line 29: In general it's good to avoid direct floating-point comparisons as they can be inaccurate. If you debug print more complicated points, you'll notice a lot of them are _close_ to 0.0, but not exactly 0.0. So most operations involving floating-point comparisons will use a "tolerance" value or threshold to see if they're "close enough".

But in this case we can go with a different approach:

The "query" API of this crate is not very fleshed out yet, but we do have the ability to search for faces that are at the extremes of certain distances. It's used in the keycap example, and using it here would look like this:

```rust
let top_edges: Vec<_> = chamfered_shape
        .faces()
        .farthest(Direction::NegZ) // Get the face whose center of mass is the farthest in the negative Z direction
        .expect("Should have a face on the bottom of the handle")
        .edges() // Get all the edges of this face
        .collect();
```

You'll notice the `edges()` function isn't yet implemented for `Face`, but it's quite simple:

```rust
pub fn edges(&self) -> EdgeIterator {
    let explorer =
        TopExp_Explorer_ctor(cast_face_to_shape(&self.inner), TopAbs_ShapeEnum::TopAbs_EDGE);

    EdgeIterator { explorer }
}
```

crates/opencascade/examples/chamfer.rs
```diff
@@ -0,0 +1,47 @@
...
+    chamfered_shape.chamfer_edges(1.0, &top_edges);
+
+    // chamfer the handle join edges
+    let handle_join_edges = chamfered_shape
+        .edges()
+        .filter(|edge| {
+            let start = edge.start_point();
+            let end = edge.end_point();
+            (start.x.abs() == 2.5 && start.z == 0.0) || (end.y.abs() == 2.5 && end.z == 0.0)
+        })
```
bschwind, line 39: Similar situation here, if we continued with this approach, we should definitely move the `5.0` into a variable (let's call it `handle_thickness`), and then use `handle_thickness / 2.0` here so we don't have to update this code if that thickness changes. We would also want to use the threshold concept I mentioned above.

_But_, OpenCascade should actually be able to give us the list of edges that were generated when running a boolean operation. You'll notice the `subtract` function actually does this already. I tried to use that same concept for the `union` operation but for some reason it's giving me zero edges. Maybe I was doing something wrong though, I'm not sure.

If you want to play around with it a bit though, it would certainly make this snippet of code more concise and straightforward.

Here's a C++ example where a Fuse operation is performed and then the newly created edges are fetched:

https://neweopencascade.wordpress.com/2018/10/17/lets-talk-about-fillets/

crates/opencascade/src/primitives.rs
```diff
@@ -372,6 +455,74 @@ impl Face {
...
+        // add all vertices from the face
+        let face_shape = cast_face_to_shape(&self.inner);
+
+        // use BRepFilletAPI_MakeFillet2d for 2d face
+        let mut make_fillet = BRepFilletAPI_MakeFillet2d_ctor(&self.inner);
+        let mut edge_map = new_indexed_map_of_shape();
+        map_shapes(face_shape, TopAbs_ShapeEnum::TopAbs_EDGE, edge_map.pin_mut());
+
+        // chamfer at vertex of all edges
+        for i in 1..=edge_map.Extent() - 1 {
```
PabloMansanet, line 499: nit:
```suggestion
        for i in 1..edge_map.Extent() {
```

PabloMansanet:
Looking great! I left a couple style nits as I haven't had time to wrap my head around the logic yet, but will look into this with more detail tomorrow 🙂

crates/opencascade/src/primitives.rs
```diff
@@ -219,21 +268,42 @@ impl Wire {
...
         let face_shape = cast_face_to_shape(&face.inner);
 
         // We use a shape map here to avoid duplicates.
-        let mut shape_map = new_indexed_map_of_shape();
-        map_shapes(face_shape, TopAbs_ShapeEnum::TopAbs_VERTEX, shape_map.pin_mut());
-
-        for i in 1..=shape_map.Extent() {
-            let vertex = TopoDS_cast_to_vertex(shape_map.FindKey(i));
-            BRepFilletAPI_MakeFillet2d_add_fillet(make_fillet.pin_mut(), vertex, radius);
+        let mut vector_map = new_indexed_map_of_shape();
```
bschwind, line 289: Should this be named `vertex_map`?

crates/opencascade/src/primitives.rs
```diff
@@ -674,6 +834,30 @@ impl Shape {
...
+
+        let chamfered_shape = make_chamfer.pin_mut().Shape();
+
+        self.inner = TopoDS_Shape_to_owned(chamfered_shape);
+    }
+
+    /// Performs fillet of `radius` on all edges of the shape
+    pub fn fillet(&mut self, radius: f64) {
+        let all_edges = self.edges().collect::<Vec<_>>();
+        self.fillet_edges(radius, &all_edges);
```
bschwind, line 852: @strohel I was running into this issue as well and I forget the best way to address this in Rust:

`edges()` returns an Iterator of owned Edges, but `chamfer_edges()` accepts an Iterator of borrowed Edges. Is there a better way to structure the function signatures so it can accept code like this?

```rust
self.chamfer_edges(distance, self.edges());
```

Basically trying to avoid the intermediate `Vec`.

crates/opencascade/src/primitives.rs
```diff
@@ -372,6 +455,74 @@ impl Face {
...
+
+        // use BRepFilletAPI_MakeFillet2d for 2d face
+        let mut make_fillet = BRepFilletAPI_MakeFillet2d_ctor(&self.inner);
+        let mut edge_map = new_indexed_map_of_shape();
+        map_shapes(face_shape, TopAbs_ShapeEnum::TopAbs_EDGE, edge_map.pin_mut());
+
+        // chamfer at vertex of all edges
+        for i in 1..=edge_map.Extent() - 1 {
+            let edge1 = TopoDS_cast_to_edge(edge_map.FindKey(i));
+            let edge2 = TopoDS_cast_to_edge(edge_map.FindKey(i + 1));
```
bschwind, line 501: From what I understand, the `edge_map` will remove duplicates, but does it guarantee any sort of order in the map? If not, grabbing `edge_1` and `edge_2` in the map as consecutive entries doesn't necessarily mean they're connected edges.

From what I saw of the cadquery code, we need to ask opencascade to give us a mapping of `Vertex` -> (`Edge`, `Edge)`, where the two edges are connected by the given `Vertex`.

crates/opencascade/src/primitives.rs
```diff
@@ -219,21 +268,42 @@ impl Wire {
...
-            let vertex = TopoDS_cast_to_vertex(shape_map.FindKey(i));
-            BRepFilletAPI_MakeFillet2d_add_fillet(make_fillet.pin_mut(), vertex, radius);
+        let mut vector_map = new_indexed_map_of_shape();
+        map_shapes(face_shape, TopAbs_ShapeEnum::TopAbs_VERTEX, vector_map.pin_mut());
+
+        let mut edge_map = new_indexed_map_of_shape();
+        map_shapes(face_shape, TopAbs_ShapeEnum::TopAbs_EDGE, edge_map.pin_mut());
+
+        // walk vertices - will chamfer edges connected to the vertex
+        // TODO should be a dict lookup of vertex to edge? Only works if both equal length - a enclosed shape
```
bschwind, line 296: Yep, I think the map is only removing duplicates, we will likely have to query for connectivity information from opencascade.

bschwind (commented):
Nice! I'm still unsure about the angled chamfers, if we can't get that working nicely soon then it's probably better to drop that from the API for now and simply have chamfer functions that just accept a distance as we know that works well.

The part I'm most unsure about is the usage of shape maps for the chamfering code. I think it works for filleting because we're just getting a unique set of vertices and operating on them one-by-one, but for edges, we need to operate on consecutive edges, and a map probably doesn't guarantee ordering.

crates/opencascade/src/primitives.rs
```diff
@@ -31,6 +33,53 @@ use opencascade_sys::ffi::{
...
+            Self::Radians(r) => (r * 180.0) / std::f64::consts::PI,
+            Self::Degrees(d) => *d,
+        }
+    }
+}
+
+pub trait ToAngle {
+    fn degrees(&self) -> Angle;
+    fn radians(&self) -> Angle;
+}
```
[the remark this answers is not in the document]
bschwind: The nice part of this trait is being able to write:

```rust
45.degrees();
```

So it's probably still useful to have, unless you have another way to accomplish that same syntax.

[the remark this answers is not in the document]
PabloMansanet: @bschwind `Degrees(45)`? 😄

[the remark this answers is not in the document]
strohel: I do see `45.degrees()` has some elegance, though I think it is a bit more Rusty/canonical to use `Degrees(45)`. Ability to name the dedicated type can also be more powerful (and complex: one then needs to decide between compile-time `Into<Radians>` vs. some runtime `enum Angle { Degrees(Degrees), Radians(Radians) }`).

tuna-f1sh:
I added the https://dev.opencascade.org/doc/occt-6.9.1/refman/html/class_top_tools___indexed_data_map_of_shape_list_of_shape.html so the edges should definitely be connecting edges. The last one is still opposite if asymmetric though 🤦

crates/opencascade/src/primitives.rs
strohel: @bschwind yeah this one is tricky in Rust.

Basically, the `Iterator` trait in `std` is written in a way that it is not possible for the yielded items to borrow from the _iterator_ itself [1]. Think of an iterator that generates the items on the fly. It is easy to create a reference for each single item yielded - but not all of them simultaneously: one would need to store them somewhere (that's `collect()`).

No solution on the iterator side, but multiple possible solutions on the `chamfer_edges()` signature side. I think the canonical solution would be to make change its signature from accepting `impl IntoIterator<Item = &'a Edge>` to accepting `impl IntoIterator<Item = T>` where `T` implements some `EdgeTrait`. `EdgeTrait` would then be implemented for both `Edge` and `&Edge`. This compiles:
```diff
diff --git a/crates/opencascade/src/primitives.rs b/crates/opencascade/src/primitives.rs
index 941a1d4..c34133d 100644
--- a/crates/opencascade/src/primitives.rs
+++ b/crates/opencascade/src/primitives.rs
@@ -180,6 +180,22 @@ impl Edge {
     pub fn tangent_arc(_p1: DVec3, _tangent: DVec3, _p3: DVec3) {}
 }
 
+pub trait EdgeTrait {
+    fn inner(&self) -> &TopoDS_Edge;
+}
+
+impl EdgeTrait for Edge {
+    fn inner(&self) -> &TopoDS_Edge {
+        &self.inner
+    }
+}
+
+impl EdgeTrait for &Edge {
+    fn inner(&self) -> &TopoDS_Edge {
+        &self.inner
+    }
+}
+
 pub struct ApproximationSegmentIterator {
     count: usize,
     approximator: UniquePtr<GCPnts_TangentialDeflection>,
@@ -822,11 +838,11 @@ impl Shape {
         self.inner = TopoDS_Shape_to_owned(chamfered_shape);
     }
 
-    pub fn fillet_edges<'a>(&mut self, radius: f64, edges: impl IntoIterator<Item = &'a Edge>) {
+    pub fn fillet_edges<T: EdgeTrait>(&mut self, radius: f64, edges: impl IntoIterator<Item = T>) {
         let mut make_fillet = BRepFilletAPI_MakeFillet_ctor(&self.inner);
 
         for edge in edges.into_iter() {
-            make_fillet.pin_mut().add_edge(radius, &edge.inner);
+            make_fillet.pin_mut().add_edge(radius, edge.inner());
         }
 
         let filleted_shape = make_fillet.pin_mut().Shape();
@@ -849,7 +865,11 @@ impl Shape {
     /// Performs fillet of `radius` on all edges of the shape
     pub fn fillet(&mut self, radius: f64) {
         let all_edges = self.edges().collect::<Vec<_>>();
+        // operate on iterator of `&Edge`
         self.fillet_edges(radius, &all_edges);
+
+        // but also on iterator of `Edge`.
+        self.fillet_edges(radius, self.edges());
     }
 
     /// Performs chamfer of `distance` on all edges of the shape
```

But it would be cumbersome to introduce something like `EdgeTrait` for each primitive type, and it also exposes internal details.

Fortunately there's a generic trait that we can use for the purpose: [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html). With that it becomes (compile-tested):
```diff
diff --git a/crates/opencascade/src/primitives.rs b/crates/opencascade/src/primitives.rs
index 941a1d4..6278126 100644
--- a/crates/opencascade/src/primitives.rs
+++ b/crates/opencascade/src/primitives.rs
@@ -180,6 +180,14 @@ impl Edge {
     pub fn tangent_arc(_p1: DVec3, _tangent: DVec3, _p3: DVec3) {}
 }
 
+// We need to provide this trivial impl: `AsRef` doesn't have this as a blanket implementation for
+// technical reasons: https://doc.rust-lang.org/std/convert/trait.AsRef.html#reflexivity
+impl AsRef<Edge> for Edge {
+    fn as_ref(&self) -> &Edge {
+        self
+    }
+}
+
 pub struct ApproximationSegmentIterator {
     count: usize,
     approximator: UniquePtr<GCPnts_TangentialDeflection>,
@@ -822,11 +830,15 @@ impl Shape {
         self.inner = TopoDS_Shape_to_owned(chamfered_shape);
     }
 
-    pub fn fillet_edges<'a>(&mut self, radius: f64, edges: impl IntoIterator<Item = &'a Edge>) {
+    pub fn fillet_edges<T: AsRef<Edge>>(
+        &mut self,
+        radius: f64,
+        edges: impl IntoIterator<Item = T>,
+    ) {
         let mut make_fillet = BRepFilletAPI_MakeFillet_ctor(&self.inner);
 
         for edge in edges.into_iter() {
-            make_fillet.pin_mut().add_edge(radius, &edge.inner);
+            make_fillet.pin_mut().add_edge(radius, &edge.as_ref().inner);
         }
 
         let filleted_shape = make_fillet.pin_mut().Shape();
@@ -849,7 +861,11 @@ impl Shape {
     /// Performs fillet of `radius` on all edges of the shape
     pub fn fillet(&mut self, radius: f64) {
         let all_edges = self.edges().collect::<Vec<_>>();
+        // operate on iterator of `&Edge`
         self.fillet_edges(radius, &all_edges);
+
+        // but also on iterator of `Edge`.
+        self.fillet_edges(radius, self.edges());
     }
 
     /// Performs chamfer of `distance` on all edges of the shape
```

That's rather neat, so I lean towards going this route.

**TL;DR:** If you need to accept both `T` and `&T` (especially with iterators and other not-simple-to-convert cases), accept `AsRef<T>`. But be sure that `impl AsRef<T> for T` exists - `stdlib` doesn't provide it.

[1] But they can borrow from some other value transitively - i.e. if the iterator already borrows from something (has a lifetime parameter), the yielded items can have that lifetime parameter. That's why `my_vec.iter()` which yields `&T` can work.

bschwind: The `AsRef` solution is perfect, thanks! I actually tried `Iterator<Item = AsRef<Edge>>` last night and for some reason my brain completely forgot about generic bounds.

strohel: I tried nested `impl` keywords `edge: impl IntoIterator<Item: impl AsRef<Edge>>`, also no avail. :shrug:

bschwind:
Whoa! I was running the keycap example to make sure everything works, and I noticed you chamfered the bottom parts of the post (or stem? whatever you call it). That was a cool surprise :D

Screen Shot 2023-07-14 at 11 18 49 AM

crates/opencascade/examples/chamfer.rs
```diff
@@ -0,0 +1,39 @@
...
+    base.chamfer(2.0);
+
+    let mut top = Workplane::xy().rect(10.0, 10.0);
+    top.translate(dvec3(0.0, 0.0, 10.0));
+    top.chamfer(1.0);
+
+    let chamfered_box = Solid::loft([&base, &top].into_iter()).to_shape();
+
+    // insert the workplane into the chamfered box area so union returns edges
+    let handle = Workplane::xy().translated(dvec3(0.0, 0.0, 0.1)).rect(5.0, 5.0);
```
bschwind, line 19: In the end I guess we're still not able to get unioned edges properly when the faces are in the same plane?

crates/opencascade-sys/src/lib.rs
```diff
@@ -32,6 +32,14 @@ pub mod ffi {
         IFSelect_RetStop,
     }
 
+    #[derive(Debug)]
+    #[repr(u32)]
+    pub enum BOPAlgo_GlueEnum {
```
bschwind, line 37: I found a good [article](https://dev.opencascade.org/content/gluing-options-boolean-component) talking about this a bit.

crates/opencascade-sys/src/lib.rs
```diff
@@ -576,6 +635,8 @@ pub mod ffi {
         pub fn Shape(self: Pin<&mut BRepAlgoAPI_Fuse>) -> &TopoDS_Shape;
         pub fn Build(self: Pin<&mut BRepAlgoAPI_Fuse>, progress: &Message_ProgressRange);
         pub fn IsDone(self: &BRepAlgoAPI_Fuse) -> bool;
+        pub fn SectionEdges(self: Pin<&mut BRepAlgoAPI_Fuse>) -> &TopTools_ListOfShape;
+        pub fn SetGlue(self: Pin<&mut BRepAlgoAPI_Fuse>, glue: BOPAlgo_GlueEnum);
```
bschwind, line 639: Were you able to experiment with this and get your handle shape working as intended?

crates/opencascade/src/primitives.rs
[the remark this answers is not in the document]
bschwind: I think for now let's just go with the simple implementation we have here, we can improve it later (create individual structs, add `const fn` support, etc.)

crates/opencascade/src/primitives.rs
```diff
@@ -535,6 +603,19 @@ impl Shape {
         self.inner = TopoDS_Shape_to_owned(filleted_shape);
     }
 
+    pub fn chamfer_edges<'a>(&mut self, distance: f64, edges: impl IntoIterator<Item = &'a Edge>) {
```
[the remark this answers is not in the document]
bschwind: Seems you can't have anonymous lifetimes in `impl Trait`, but maybe this will go away anyway when we move to your `AsRef` solution.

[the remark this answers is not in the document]
strohel: Ah, thanks for checking.

bschwind (approved):
I'd say this is good to merge once CI passes! I pushed a few commits to address some small nitpicks and code/comment formatting. Thanks for this nice addition to the code!

PR #79, opened 2023-07-12
Add missing workplanes and affine transforms (plus bonus gizmo geometry)

PabloMansanet:
The point of this PR is just to get the missing affine transforms in the main branch quickly, but since the `viewer` main does have a bit of a soup of geometries already, I figured I'd also throw in the gizmo in progress (will use that shape as a basis for the actual gizmo behaviour).

### DIFF
--- a/crates/opencascade/src/workplane.rs
+++ b/crates/opencascade/src/workplane.rs
@@ -37,69 +37,48 @@ impl Plane {
                 dvec3(1.0, 0.0, 0.0),
                 dvec3(0.0, 0.0, 0.0),
             ),
-            Self::ZX => {
-                // TODO - fix this
-                DAffine3::from_cols(
-                    dvec3(1.0, 0.0, 0.0),
-                    dvec3(0.0, 1.0, 0.0),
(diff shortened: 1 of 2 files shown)


crates/opencascade/src/workplane.rs
```diff
@@ -25,106 +25,22 @@ impl Plane {
...
-            Self::Bottom => {
-                // TODO - fix this
-                DAffine3::from_cols(
-                    dvec3(1.0, 0.0, 0.0),
-                    dvec3(0.0, 1.0, 0.0),
-                    dvec3(0.0, 0.0, 1.0),
-                    dvec3(0.0, 0.0, 0.0),
-                )
-            },
+            Self::XY => DAffine3::from_cols(DVec3::X, DVec3::Y, DVec3::Z, DVec3::ZERO),
```
bschwind, line 28: For any reviewers, here's a [reference](https://cadquery.readthedocs.io/en/latest/classreference.html?highlight=named#cadquery.Plane.named)

bschwind (approved):
LGTM! Good to merge once CI is green

PR #81, opened 2023-07-14
Coalesce `Shape`s, move ad-hoc API to `AdHocShape` wrapper

strohel:
I was initially confused why there are two `Shape` struct definitions (with similar payload), and why their methods partially overlap, but with differences in the signature.

This PR doesn't resolve the duplication, but should make it clear what is deprecated and what is recommended, and what kind of API are consumers using.

It also makes Shape and AdHocShape fully interchangeable: AdHocShape can be created from Shape, destructures and (mutably) dereferences to it. That should make it possible to port usages incrementally.

Locally this diff is best viewed with `-C` / `--find-copies` to detect the ad-hoc API was just moved.

---
My initial try was to make the ad-hoc API an extension trait for `primitive::Shape`, but that didn't work well: many methods alias between the primitive and ad-hoc API, so users of the ad-hoc API would often have to use the ugly fully-qualified trait call syntax.


bschwind:
Sorry for the merge conflict, the imports path from `opencascade-sys` is a merge-conflict magnet.

bschwind (approved):
LGTM! Just needs the merge conflict fixed.

strohel:
Rebased, I'll merge on CI is green after 30 minutes. :upside_down_face:

> Sorry for the merge conflict, the imports path from opencascade-sys is a merge-conflict magnet.

Fortunately conflicts in imports are super-easy to resolve - I just click `Accept both changes` in VSCode and then let rust-analyze deduplicate the imports/remove unused ones.

But the import block still looks funny (mildly speaking). I thought we could import just
```rust
use use opencascade_sys::ffi;
```
and then call the methods as `ffi::gp_Ax2_ctor()`? If that sounds good, I can fire a quick PR with that.

bschwind:
> and then call the methods as ffi::gp_Ax2_ctor()? If that sounds good, I can fire a quick PR with that.

Oh! So simple and obvious, I'm totally in favor of that :)

PR #83, opened 2023-07-15
Iterator and other API quality-of-life improvements

bschwind:
This applies the [wonderful suggestion](https://github.com/bschwind/opencascade-rs/pull/76#discussion_r1261713852) by @strohel for accepting iterators which give either `T` or `&T`, making for cleaner code at the call site (no need for intermediate calls to `.collect()`)

This also changes the `farthest()` function to return `T` instead of  `Option<T>` because most model code will be immediately unwrapping anyway. And if you don't have a face that is farthest in some direction, there's probably something wrong with the model. I actually don't know what the expected outcome is if there are multiple faces that are _exactly_ the same distance along some axis, I should check cadquery to see how it handles that.

### DIFF
--- a/crates/opencascade/examples/chamfer.rs
+++ b/crates/opencascade/examples/chamfer.rs
@@ -27,7 +27,6 @@ pub fn main() {
     let top_edges = chamfered_shape
         .faces()
         .farthest(Direction::NegZ) // Get the face whose center of mass is the farthest in the negative Z direction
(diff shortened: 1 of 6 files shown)


strohel (approved):
Oh nice! I didn't realize we had to collect _that_ often.

Supernit suggestion: add an explaining comment on the first occurrence of `impl AsRef<T> for T`.

PR #85, opened 2023-07-15
Add an Angle and RVec (rotation vector) type

bschwind:
This is _one_ possible implementation for more strongly-typed angles. Spawned from [this discussion](https://github.com/bschwind/opencascade-rs/pull/76#discussion_r1261175426)

The main goals are:

* Calling code should always indicate the unit of rotation (radians vs. degrees vs. something else)
* The called code (the OpenCascade bindings) should also explicitly mention which rotation unit it desires when passing an angle to OpenCascade.
* The calling code should be ergonomic (easy to construct, easy to read, terse but not to the point where it hurts readability)

I think what we have here so far is nice, but I'm open to suggestions if we want to make it more Rust-y with new types and `impl From<_>` signatures on functions which take angles.

### DIFF
--- a/crates/opencascade/examples/keycap.rs
+++ b/crates/opencascade/examples/keycap.rs
@@ -2,6 +2,7 @@
 // https://github.com/cubiq/OPK/blob/53f9d6a4123b0f309f87158115c83d19811b3484/opk.py
 use glam::dvec3;
 use opencascade::{
+    angle::{RVec, ToAngle},
     primitives::{Direction, Face, Solid},
     workplane::Workplane,
 };
@@ -13,7 +14,7 @@ pub fn main() {
     let keycap_unit_size_x = 1.0;
     let keycap_unit_size_y = 1.0;
     let height = 16.0;
-    let angle = 13.0;
+    let angle = 13.0.degrees();
     let depth: f64 = 2.8;
     let thickness: f64 = 1.5;
     let base = 18.2;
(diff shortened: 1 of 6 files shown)


strohel:
@PabloMansanet can we nerd-snipe you to try an alternative implementation where `Degrees` and `Radians` are separate types?

I _think_ the "2 separate types" would be just slightly more canonical, yet a bit more verbose: APIs would have to accept `argument: impl Into<Radians>` (which we could alias as `impl Angle`) rather than `argument: Angle`.

My gut feeling would be `enum { Degrees, Radians }` would be more suitable when we want to carry the difference between the 2 variants in runtime. But if we only care about interoperability, compile-time conversion (the 2 separate types) is the way to go.

But both approaches are better than status quo, I'm certainly not _against_ merging this.

PabloMansanet:
> @PabloMansanet can we nerd-snipe you to try an alternative implementation where `Degrees` and `Radians` are separate types?
> 
> I _think_ the "2 separate types" would be just slightly more canonical, yet a bit more verbose: APIs would have to accept `argument: impl Into<Radians>` (which we could alias as `impl Angle`) rather than `argument: Angle`.
> 
> My gut feeling would be `enum { Degrees, Radians }` would be more suitable when we want to carry the difference between the 2 variants in runtime. But if we only care about interoperability, compile-time conversion (the 2 separate types) is the way to go.
> 
> But both approaches are better than status quo, I'm certainly not _against_ merging this.

I like being nerd sniped :) sure thing, I'll get through this today, even if I don't manage to get much more done other than review!

PR #88, opened 2023-07-18
Add revolve feature to face primitives

### LINKED ISSUE #41: Support revolutions (rotating a profile around an axis to create a solid)

katyo:
Closes #41

### DIFF
--- a/crates/opencascade/src/primitives.rs
+++ b/crates/opencascade/src/primitives.rs
@@ -20,6 +20,10 @@ pub fn make_vec(vec: DVec3) -> UniquePtr<ffi::gp_Vec> {
     ffi::new_vec(vec.x, vec.y, vec.z)
 }
 
+pub fn make_axis_1(origin: DVec3, dir: DVec3) -> UniquePtr<ffi::gp_Ax1> {
+    ffi::gp_Ax1_ctor(&make_point(origin), &make_dir(dir))
+}
+
 pub fn make_axis_2(origin: DVec3, dir: DVec3) -> UniquePtr<ffi::gp_Ax2> {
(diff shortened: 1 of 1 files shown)


bschwind:
@katyo FYI the usage of `Angle` is currently being added in #85

crates/opencascade/src/primitives.rs
```diff
@@ -425,6 +429,21 @@ impl Face {
         Shape { inner }
     }
 
+    pub fn revolve(&self, origin: DVec3, axis: DVec3, angle: Angle) -> Solid {
```
bschwind, line 432: I'd like to suggest we make two functions

```
revolve((&self, origin: DVec3, axis: DVec3)
```

and

```
revolve_angle((&self, origin: DVec3, axis: DVec3, angle: Angle)
```

Full revolves feel like they're done often enough that it would be annoying to always have to specify the angle.

bschwind:
@katyo You'll likely want to rebase on the latest `main`, the `Angle` change in #85 has been merged.

crates/opencascade/src/primitives.rs
katyo: I dont like have two functions for similar things. May be better turn angle parameter to be Option.

bschwind: Fair enough, the `Option<Angle>` API looks good.

I _might_ tweak it later as I observe usage. If there's a lot of `shape.revolve(origin, axis, None)` everywhere I might split out the angled version to a separate function. Nothing to worry about for this PR though.

bschwind (approved):
LGTM, thanks for the PR!

PR #91, opened 2023-07-19
Start work on excluding unnecessary CMake modules

bschwind:
Goals of this PR:

* Reduce the size of the `occt-sys` crate to below 10 MB, the crates.io upload limit
* Reduce the amount of work CMake needs to do to build opencascade

I realized we can override opencascade files with the CMake `BUILD_PATCH` variable. I recreated the `adm/MODULES` directory in our patch directory and verified that CMake picks it up and uses it instead of the file in OCCT.

Next step is to modify `adm/MODULES` to cut out things we don't need, as well as editing `occt-sys/Cargo.toml` to exclude big directories that we don't use anyway.

Here's a [reference](https://dev.opencascade.org/doc/overview/html/build_upgrade__building_occt.html#build_occt_win_cmake) to some of the CMake stuff in occt.

Depends on #112 being merged first.


crates/occt-sys/build.rs
```diff
@@ -2,19 +2,31 @@ const LIB_DIR: &str = "lib";
 const INCLUDE_DIR: &str = "include";
 
 fn main() {
+    let current_dir = std::env::current_dir().expect("Should have a 'current' directory");
+    let patch_dir = current_dir.join("patch");
+
     let dst = cmake::Config::new("OCCT")
+        // TODO(bschwind) - Verify this is set properly
+        // .define("CMAKE_BUILD_TYPE", "Release")
```
strohel, line 10: CMAKE_BUILD_TYPE should be handled correctly by `cmake` crate, ref. https://docs.rs/cmake/latest/cmake/struct.Config.html#method.profile

bschwind: Ah nice, I had hoped that was the case but hadn't checked yet.

strohel:
I realize that this could *also* be a work-around for CI caching. Once we publish `occt-sys` to crates.io, we can depend on it through crates.io, including it in the cache (hopefully).

bschwind:
> Once we publish occt-sys to crates.io, we can depend on it through crates.io, including it in the cache (hopefully).

Ah yeah, if that works well then I'd much prefer that approach.

bschwind:
Trying hard to cut out all unnecessary files to keep the crate size under the 10MB limit - I think we may need a different approach of splitting up toolkits into different crates. I've gotten fairly good at controlling which "toolkits" (as occt calls them) get built, but getting them all to fit in one crate is looking very difficult and time consuming to accomplish.

strohel:
Issues linked to https://github.com/rust-lang/crates.io/issues/195 seems to be like a good library of workarounds people do to circumvent the limit.

- Perhaps we can do something similar to https://github.com/blas-lapack-rs/openblas-src/pull/95 ?

bschwind:
cc @mkovaxx

bschwind:
@strohel if we specify `occt-sys` as a git dependency, will our cargo cache action cache it properly? I think we wouldn't be able to publish on crates.io with a git dependency, but maybe it would help with the caching situation?

bschwind:
Merging this for now, there can be some cleanup afterwards but I want to get this in now to get CI build times down.

strohel (commented):
Nice!

Issue #61, opened 2023-07-03
Improve camera movement in the current app


bschwind:

bschwind:
Here's a good reference on CAD cameras:

https://elalish.blogspot.com/2022/04/3d-interaction.html

PR #93, opened 2023-07-19
Support orthographic projection and a key toggle

skywhale:
For https://github.com/bschwind/opencascade-rs/issues/61

### DIFF
--- a/crates/viewer/src/camera.rs
+++ b/crates/viewer/src/camera.rs
@@ -31,6 +31,9 @@ impl Camera {
     }
 
     pub fn matrix(&self) -> Mat4 {
+        // These magic numbers are configured so that the particular model we are loading is
+        // visible in its entirety. They will be dynamically computed eventually when we have "fit
+        // to view" function or alike.
         let proj = match self.projection {
             Projection::Orthographic => Mat4::orthographic_rh(
                 -100.0 * self.aspect_ratio,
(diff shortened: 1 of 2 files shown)


crates/viewer/src/camera.rs
```diff
@@ -0,0 +1,56 @@
...
+                100.0 * self.aspect_ratio,
+                -100.0,
+                100.0,
+                -1000.0,
+                1000.0,
+            ),
+            Projection::Perspective => {
+                Mat4::perspective_rh(std::f32::consts::PI / 2.0, self.aspect_ratio, 0.01, 1000.0)
+            },
+        };
```
strohel, line 46: I see that it is preexisting, but where the 1000 (and 100) constant comes from?

strohel (approved):
Nice! Just a curiosity question (independent from merging)

tuna-f1sh (approved):
Checked it out and works as I would expect. Switching between the two with the hidden lines visible it is like the infinite stairs illusion - I thought it was flipping at first!

crates/viewer/src/camera.rs
skywhale: This is a random constatnt so we can see the wholeness of the particular model we are loading. Once we have "scale to fit" function, we should be able to get rid of it and compute it more dynamically. Let me add a documentation.

skywhale:
Thank you for the review @tuna-f1sh @strohel :) Added a clarying comment and merging.

bschwind:
@skywhale I'd like to review it too if you can wait a bit before merging

skywhale:
@bschwind no problem :) thank you!

bschwind (approved):
LGTM! I was going to suggest caching the camera matrix instead of just the aspect ratio but the performance overhead is so trivial it's not worth doing. And we used to calculate it every frame before anyway.

Issue #59, opened 2023-06-30
Support KiCAD import and export


bschwind:

tuna-f1sh:
Good points. My initial thought process, considering how I approach this manually:

* I also thought about how to get part clearances. Not just the connectors but one also needs to know the top and bottom clearance required; the highest part + board thickness/2 + clearance. Either we try to extract this from STEP or probably better: a flag or [board variable](https://docs.kicad.org/master/id/eeschema/eeschema_advanced.html)
* A clearance variable is probably sensible to be defined for the board outline, top and bottom. Again a flag at export or in the .kicad_pcb file.
* For the mounting holes, I guess we just start with a basic stand-off style extruded from the screw hole to where the courtyard finishes.
* Here is my recent project with some nice _challenging_ IO...the USB-C connector, plug and SWD header on the top. Not too challenging though.

bschwind:
> Either we try to extract this from STEP or probably better: a flag or [board variable](https://docs.kicad.org/master/id/eeschema/eeschema_advanced.html)

Oh nice, I didn't know about board variables! That seems like it may be the most straightforward approach. If we can set this up so it's more self-contained in the kicad project, I think it introduces less friction than having to export a particular file from kicad every time.

> For the mounting holes, I guess we just start with a basic stand-off style extruded from the screw hole to where the courtyard finishes.

Yep, let's start with something simple like that for now.

> Here is my recent project with some nice challenging IO...the USB-C connector, plug and SWD header on the top. Not too challenging though.

Oh that's a good sample project, I hadn't even thought of connectors perpendicular to the board but of course those are quite common.

tuna-f1sh:
I did a bit of a review of any existing KiCad Rust helpers since I think a good first start is to make a box around the board 'Edge.Cuts'. It looks like this:

```
  (gr_line (start 185 71) (end 151 71)
    (stroke (width 0.1) (type default)) (layer "Edge.Cuts") (tstamp 0d0b18b8-29e1-400d-9fe0-0c553320d369))
  (gr_line (start 188.950546 53.08) (end 188.950546 47.524454)
    (stroke (width 0.1) (type default)) (layer "Edge.Cuts") (tstamp 19f9a363-e864-4747-9205-86bc18fec056))
  (gr_arc (start 188.675546 47.249454) (mid 188.87 47.33) (end 188.950546 47.524454)
[...]

PR #94, opened 2023-07-19
Add the beginning of a very simplistic kicad board file parser

bschwind:
Relates to #59

This is a _very_ naively written parser, which is quite specific and targeted to just extracting the info we want from the boards. It's much more manual work. I wouldn't view this as a permanent solution, but more of a short-term thing to get the data we want.

Currently just graphic lines (`gr_line`) are extracted, but the code is structured in a way that it's fairly easy to add more extractors.

Ideally, the kicad parser would integrate with `serde` or work with streaming data for larger files (probably not necessary for now).

This PR is mostly just for discussion and to see if we'd want to go this direction for short-term file parsing.


tuna-f1sh:
I checked this out and it does seem like a nice simple option - since we only _need_ the `gr_*` parts. It also feels easier to add the other sections as required. 

Have you thought about the next step? I had a quick look at .DXF format and dxf crate since it looks like there is a lot cross-over in 2d shapes: https://docs.rs/dxf/latest/dxf/entities/index.html. 

But perhaps it is just as simple a `to_wire` for a passed layer in the `KicadBoard`, perhaps with an enum for a `BoardLayer`:

```
pub enum BoardLayer {
    FCu,
    BCu,
    FAdhes,
    BAdhes,
    FPaste,
    BPaste,
    FSilkS,
    BSilkS,
    FMask,
    BFask,
    DwgsUser,
    CmtsUser,
    Eco1User,
    Eco2User,
    EdgeCuts,
    Margin,
    BCrtYd,
    FCrtYd,
    BFab,
    FFab,
    In1Cu,
    In2Cu,
    In3Cu,
    In4Cu,
    User(String)
}
```

tuna-f1sh:
@bschwind had chance to play with this a bit today. I've implemented the `gr_*` types that match our primitives. Some `Into` for converting them to primitives and some methods to the `KicadBoard` to get these for a `BoardLayer`. Also split stuff into files since it was getting a bit busy - hope that's okay.

I think that the `Graphic*` types could perhaps have some elements made generic since there is some duplication. Maybe not though.

The result is a .stl file output...well very simple shape one (the sample.kicad_pcb does not work). Basically I can use Kicad as a simple sketching tool 😄; rectangles, circles and lines . What isn't working is `GraphicLine` and `GraphicArc` combinations - perhaps what I am doing here is at fault: https://github.com/bschwind/opencascade-rs/pull/94/commits/cc726da5b19c504276ca73a1eddae549ab08af3f#diff-4ada67e285da4eb727355d07da45ba9924bc9608b44873a85d4a818dc9694a9fR184

Some quick testing, it seems like them being out of order causes issue - but there is no guarantee they will be in order in the file. Not 100% this is the issue but here is an example:

# Works
Screenshot 2023-07-22 at 17 30 47

Screenshot 2023-07-22 at 17 30 51

# Not closed properly
Screenshot 2023-07-22 at 17 31 23

Screenshot 2023-07-22 at 17 31 27

tuna-f1sh:
@bschwind ah, I just read https://dev.opencascade.org/doc/refman/html/class_b_rep_builder_a_p_i___make_wire.html#a8a6bc22ee4601968fe013021875f06b9

> E must be connectable to the wire under construction, and, unless it is the first edge of the wire, must satisfy the following condition: one of its vertices must be geometrically coincident with one of the vertices of the wire (provided that the highest tolerance factor is assigned to the two vertices). It could also be the same vertex. Warning If E is not connectable to the wire under construction it is not added

That explains it then. So I guess we need to sort the lines and edges so this is obeyed.

tuna-f1sh:
Came back to this a slightly dirty fix. It turns out that `MakeWire` will return false for `IsDone` if the last `Edge` was not added. It's possible to use this to loop around until they are all added or the missing ones stop reducing: https://github.com/bschwind/opencascade-rs/pull/94/commits/58f779966447b4077a0c2268b68ad5441782d3fb

Not sure whether `from_edges` should provide this behavior or expect the Edges to be passed in connecting order. It should probably return a `Result` if we add this behavior. 

Screenshot 2023-07-22 at 21 13 49

bschwind:
@tuna-f1sh  I'm on my phone so I'll link better references later, but I'm pretty sure OCC has the ability to take in a set of wires and connect them properly. I think CadQuery was using that functionality under the name "AssembleEdges" or something like that.

crates/opencascade/src/primitives.rs
```diff
@@ -213,9 +213,28 @@ impl AsRef<Wire> for Wire {
...
+        // This will get updated by missing edges and exit when all edges are added
+        while edges.len() > 0 {
+            let mut missing_edges = Vec::with_capacity(edges.len());
+            let start_len = edges.len();
+
+            for edge in edges {
+                make_wire.pin_mut().add_edge(&edge.inner);
+                // If MakeWire IsDone returns false it means the edge was not added due to no shared vertices
+                // We will assume it is due to the edges being out of order and will attempt to add again
+                if make_wire.pin_mut().IsDone() != true {
```
bschwind, line 227: Ah I was close, it's `consolidateWires` in cadquery, which indeed calls [`ConnectEdgesToWires`](https://dev.opencascade.org/doc/refman/html/class_shape_analysis___free_bounds.html#a4809fe812b6a663a286df71b4d7bda3b)

crates/kicad-parser/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "kicad-parser"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1"
+sexp = "1.1.4"
+opencascade = { version = "0.1", path = "../opencascade" }
+glam = { version = "0.23", features = ["bytemuck"] }
```
bschwind, line 10: I think for the `kicad-parser` crate specifically, it can just be focused on parsing the data into nice Rust structs.

We can have another module in the `opencascade` crate specifically for generating shapes from the data this crate pulls in. So even `glam` is probably unnecessary as a dependency here, since we're only using it to construct `dvec3` from the parsed data without doing any math.

This crate can keep things in their original parsed 2D coordinates, and it can be the responsibility of the `opencascade` layer to place those 2D "sketches" in the desired workplane (Z might not be `0` for all use cases!).

I'd be happy to make this change, or stub out what I think the higher level kicad code would look like.

tuna-f1sh: Ah okay makes sense. So remove all the `Into` primitives too? I guess moving those into another '2d graphics' crate means they can be used by another 2d import such as .dxf.

I'll hand it to you since it sounds like you have an idea of what you're expecting.

crates/opencascade/src/primitives.rs
tuna-f1sh: I also think this is quite nice to have in the `from_edges` - not a panic but a `Result` if all the edges are not added.

crates/kicad-parser/Cargo.toml
bschwind: > So remove all the Into primitives too?

Likely so, I think that will happen at a higher level. Sorry for being somewhat vague but I'll try to follow up on this to show what I was thinking for the design.

bschwind:
This has enough functionality to be able to import the edge cuts on a board now, including footprints which have edge cuts. I'll go ahead and merge it and refine it more later.

PR #100, opened 2023-07-23
feat/revolve: Fix angle converting

katyo:
Fixup for #88


katyo:
Seems I forget re-apply cherry picks from my upcoming when updating PR branch.
Looks like we highly need some CI to prevent at least compilation errors.

bschwind:
Hey sorry I missed this, I already pushed a fix to `main` which has the `radians()` function take `self` instead of `&self` because `Angle` implements `Copy` anyway.

bschwind:
> Looks like we highly need some CI to prevent at least compilation errors.

Hmm, I see now why CI didn't run for your branch - I configured the workflows to only run for `push`, not `pull_request`.

I had disabled `pull_request` because when _I_ push a branch and open a PR, CI runs twice. Let me see if I can fix it up so it still runs for PRs from anyone, without running twice.

PR #101, opened 2023-07-23
Run CI on pull requests, or pushes to main

bschwind:


strohel (approved):
Yeah this makes sense.

Out-of-scope thoughts:
- given all the workflows have the same `on` condition, we could make this a single workflow with multiple jobs (I'd say GitHub actions have one too many levels of abstraction)
- We could try enabling merge trains. If I understand this correctly, it runs the CI right before merging but after rebasing/squashing. That should catch things like "we've changed this method signature in main in the mean time" even if they don't introduce a merge conflict.

bschwind:
I'd be in favor of enabling merge queues. I poked around the settings but didn't see where exactly I can enable it. Is it a beta feature, or do I need to use it via .github workflow files?

strohel:
> I'd be in favor of enabling merge queues. I poked around the settings but didn't see where exactly I can enable it. Is it a beta feature, or do I need to use it via .github workflow files?

Ah, no luck this time: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue
> Pull request merge queues are available in any public repository owned by an organization, or in private repositories owned by organizations using GitHub Enterprise Cloud.

PR #103, opened 2023-07-23
RFC: viewer: draw XYZ axes

### LINKED ISSUE #62: Add XYZ grids

strohel:
This is extremely naive implementation, but firing it just to get going. It seems to work, but I cannot rotate or zoom in the viewer (that's a privilege that only Mac users have at the moment).

Once finalized, fixes #62.

Screenshot_20230723_181803

### DIFF
--- a/crates/viewer/src/main.rs
+++ b/crates/viewer/src/main.rs
@@ -83,6 +83,8 @@ impl GameApp for ViewerApp {
             line_builder.add_round_line_strip(&segments);
         }
 
+        draw_xyz_axes(&mut line_builder);
+
         let rendered_edges = line_builder.build(graphics_device.device());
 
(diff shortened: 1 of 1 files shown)


bschwind:
I found an interesting approach to drawing an infinite grid, with anti-aliasing and fade-out (sometimes called fog):

https://asliceofrendering.com/scene%20helper/2020/01/05/InfiniteGrid/

We don't need anti-aliasing as we're using `smaa-rs`, but it might be interesting to borrow some techniques from there, especially the infinite part.

bschwind:
I'll go ahead and close this for now, as I think there are some nice looking solutions to drawing an infinite grid linked in comments in the main issue for this feature.

PR #105, opened 2023-07-24
ci: Avoid unmaintained actions-rs

katyo:
Because actions-rs/toolchain and actions-rs/cargo hasn't maintained over two years we probably should avoid using it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,13 +14,9 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: "recursive"
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@v1
         with:
-          profile: minimal
           toolchain: stable
-          override: true
       - name: Cache Rust Dependencies
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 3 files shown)


bschwind (approved):
Huh, I didn't know those were unmaintained. I'm always in favor of anything dtolnay maintains :)

@strohel, what do you think?

.github/workflows/test.yml
```diff
@@ -14,13 +14,9 @@ jobs:
...
         with:
-          profile: minimal
           toolchain: stable
-          override: true
       - name: Cache Rust Dependencies
         uses: Swatinem/rust-cache@v2
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
+      - run: cargo test --all-targets
```
strohel, line 22: Curious: I wonder the the actual effect of `--all-targets` here is: I thought "tests" is one of the targets? It certainly doesn't matter having it here. I see it is documented in `cargo test --help`: `--all-targets             Test all targets`.

strohel (approved):
Also didn't know `actions-rs` are unmaintained. (we should update our other repos) Looks good!

bschwind:
Thanks, @katyo!

PR #107, opened 2023-07-24
Add a BooleanShape struct to hold the result of CSG operations

### LINKED ISSUE #48: Support easy filleting of edges when shapes are unioned or subtracted from each other
Reference: https://neweopencascade.wordpress.com/2018/10/17/lets-talk-about-fillets/

(use the `SectionEdges()` function on a boolean operation to retrieve newly formed edges)

bschwind:
Closes #48

This is still a work in progress, I also want to add your typical operator overloads for things like union and subtraction, at least.

### DIFF
--- a/crates/opencascade/src/primitives.rs
+++ b/crates/opencascade/src/primitives.rs
@@ -32,7 +32,7 @@ pub fn make_axis_2(origin: DVec3, dir: DVec3) -> UniquePtr<ffi::gp_Ax2> {
 }
 
 pub struct Vertex {
-    _inner: UniquePtr<ffi::TopoDS_Vertex>,
+    inner: UniquePtr<ffi::TopoDS_Vertex>,
 }
 
 // You'll see several of these `impl AsRef` blocks for the various primitive
@@ -57,7 +57,7 @@ impl Vertex {
         let vertex = make_vertex.pin_mut().Vertex();
         let inner = ffi::TopoDS_Vertex_to_owned(vertex);
 
-        Self { _inner: inner }
+        Self { inner }
     }
 }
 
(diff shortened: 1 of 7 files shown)


julianschuler (changes requested):
That is a really nice addition! I just have one minor remark regarding the From and `to_shape` implementations.

crates/opencascade/src/primitives.rs
```diff
@@ -807,7 +863,65 @@ impl AsRef<Shape> for Shape {
...
+impl From<Wire> for Shape {
+    fn from(wire: Wire) -> Self {
+        let shape = ffi::cast_wire_to_shape(&wire.inner);
+        let inner = ffi::TopoDS_Shape_to_owned(shape);
+
+        Shape { inner }
+    }
+}
+
+impl From<Face> for Shape {
```
julianschuler, line 893: I really like to see the those From implementations for all those primitives! However, many of the primitives also have a `to_shape` function, which could be deduplicated by using `.into()`. Even better may be to implement a trait which automatically provides `to_shape()` using the From implementations:

```rust
pub trait ToShape {
    fn to_shape(self) -> Shape;
}

impl<T: Into<Shape>> ToShape for T {
    fn to_shape(self) -> Shape {
        self.into()
    }
}
```

bschwind:
@julianschuler thanks for taking a look! Maybe you haven't published it yet but I currently don't see any other review comments.

julianschuler (changes requested):
Now also with the actual suggestion... 😅

crates/opencascade/src/primitives.rs
bschwind: Great suggestion! I felt like I still wanted to hang onto the `to_shape()` function because it seemed clearer than `.into()` with type ascription. I'll play around with your idea and get it added.

julianschuler (approved):
LGTM!

PR #108, opened 2023-07-24
Support pan, rotate and zoom camera operations

skywhale:
Rotate and zoom is achieved by moving the camera instead of modifying the target object.

For https://github.com/bschwind/opencascade-rs/issues/61

### DIFF
--- a/crates/viewer/src/camera.rs
+++ b/crates/viewer/src/camera.rs
@@ -92,7 +92,7 @@ impl OrbitCamera {
                 let proj = Mat4::perspective_rh(
                     std::f32::consts::PI / 2.0,
                     self.aspect_ratio,
-                    10.0,
+                    1.0,
                     10_000.0,
                 );
                 (proj, self.zoom_factor * self.radius)


bschwind:
Hey thanks! I took a look at this and will get back to it more later. It mostly works but there are a few edge cases and issues to work out before merging. I'll go into more detail on a review later.

tuna-f1sh:
Nice to be able to move the model at last. I also noticed probably the same oddities:

* Perhaps just me (I often invert y axis on games) but it feels reversed to what I would expect.
* The object can _clip_ (for want of a better word) zooming out - not found the culprit yet.

Screenshot 2023-07-26 at 19 03 14

skywhale:
@bschwind @tuna-f1sh I switched to use quartenion rotation and reversed the rotation orientation. PTAL 🙇 

I could not figure out what is causing the clipping issue, though. The `far` and `near` parameters that I'm passing to projection do not seem related.

crates/viewer/src/camera.rs
```diff
@@ -11,11 +13,26 @@ enum Projection {
...
     aspect_ratio: f32,
+    // Zoom factor used for orthographic projection.
+    zoom_factor: f32,
+    // Position of the camera.
+    position: Vec3,
+    // The upward vector of the camera, determining its orientation.
+    upward: Vec3,
+    // The look-at target, in the center of the view.
+    target: Vec3,
 }
```
mkovaxx, line 24: Talking to Brian, it sounds like the common case is a camera that orbits a target point.
So the behavior is likely easier to implement if the data representation is made for that.

Something like this:
```suggestion
pub struct OrbitCamera {
    projection: Projection,
    aspect_ratio: f32,
    // Zoom factor used for orthographic projection.
    zoom_factor: f32,
    // The look-at target, in the center of the view.
    target: Vec3,
    // The radius of the orbit
    radius: f32,
    // The orientation of the camera around the target point
    orientation: Quat,
}
```

Working with this representation:
- `upward` and `backward` vectors can be extracted from `orientation` on demand
- `position` can be computed on demand as `target + radius * backward`

crates/viewer/src/camera.rs
```diff
@@ -30,16 +47,45 @@ impl Camera {
...
+
+    /// Orbit around the target while keeping the distance.
+    pub fn rotate(&mut self, yaw: f32, pitch: f32) {
+        let backward = (self.position - self.target).normalize();
+        let leftward = self.upward.cross(backward);
+        let rotation =
+            Quat::from_axis_angle(self.upward, yaw) * Quat::from_axis_angle(leftward, pitch);
+        self.position = (self.target + rotation * backward) * backward.length();
+        self.upward = rotation * self.upward;
+    }
```
mkovaxx, line 77: With the representation suggested above, this could be greatly simplified:
```suggestion
    pub fn rotate(&mut self, rotator: Quat) {
        self.orientation *= rotator;
    }
```

mkovaxx (commented):
@skywhale Hey Ryo, this is Máté! Just sharing some thoughts. Hope they'll help!

crates/viewer/src/camera.rs
skywhale: Thank you @mkovaxx! This makes a lot of sense. Let me apply the changes :)

skywhale: @mkovaxx How do I extract `upward` and `backward` from `orientation`. I'm still wrapping my heads around quaternion and glam API.

mkovaxx:
I'm going to commandeer this PR! ;)

mkovaxx:
@bschwind I think this works reasonably well for now. Please try it out! ;)

mkovaxx:
Rebased onto latest `main`.

crates/viewer/src/camera.rs
```diff
@@ -8,14 +10,29 @@ enum Projection {
...
+impl OrbitCamera {
     pub fn new(width: u32, height: u32) -> Self {
-        Self { projection: Projection::Orthographic, aspect_ratio: width as f32 / height as f32 }
+        Self {
+            projection: Projection::Orthographic,
+            aspect_ratio: width as f32 / height as f32,
+            zoom_factor: 1.0,
+            target: vec3(0.0, 0.0, 0.0),
+            radius: 100.0,
+            orientation: Quat::IDENTITY,
```
bschwind, line 34: Here's my take on it:

```rust
let initial_camera_pos = vec3(2.0, -2.0, 1.0);

...

orientation: Quat::from_rotation_arc(
    Vec3::Z,
    initial_camera_pos.normalize(),
) * Quat::from_axis_angle(Vec3::Z, std::f32::consts::PI / 4.0),
```

mkovaxx: I made it parametric so you can just plug numbers in!

crates/viewer/src/camera.rs
```diff
@@ -30,29 +51,58 @@ impl Camera {
...
+    }
+
+    /// Zoom in or out, while looking at the same target.
+    pub fn zoom(&mut self, zoom_delta: f32) {
+        self.zoom_factor = f32::max(self.zoom_factor * f32::exp(zoom_delta), MIN_ZOOM_FACTOR);
+    }
+
+    /// Orbit around the target while keeping the distance.
+    pub fn rotate(&mut self, rotator: Quat) {
+        self.orientation *= rotator;
```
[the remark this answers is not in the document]
mkovaxx: It's a small concern. I'll take care of it, thanks for flagging it! :)

crates/viewer/src/camera.rs
```diff
@@ -8,14 +10,33 @@ enum Projection {
...
 
-impl Camera {
-    pub fn new(width: u32, height: u32) -> Self {
-        Self { projection: Projection::Orthographic, aspect_ratio: width as f32 / height as f32 }
+impl OrbitCamera {
+    pub fn new(width: u32, height: u32, init_pos: Vec3) -> Self {
+        let target = Vec3::ZERO;
+        let radius = init_pos.length();
+        let look_at_matrix = Mat4::look_at_rh(init_pos, target, Vec3::Z);
+        let orientation = Quat::from_mat4(&look_at_matrix).inverse();
```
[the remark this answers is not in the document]
mkovaxx: Thank you sir! ;)

crates/viewer/src/camera.rs
```diff
@@ -30,29 +51,60 @@ impl Camera {
...
+                    1000.0,
+                );
+                (proj, self.radius)
+            },
             Projection::Perspective => {
-                Mat4::perspective_rh(std::f32::consts::PI / 2.0, self.aspect_ratio, 0.01, 1000.0)
+                let proj = Mat4::perspective_rh(
+                    std::f32::consts::PI / 2.0,
+                    self.aspect_ratio,
+                    10.0,
```
bschwind, line 95: I played around a bit with this, I think `1.0` is a bit more comfortable for most of the models I tried.

bschwind, line 95:
```suggestion
                    1.0,
```

bschwind (approved):
Nice! I'll go ahead and make the change I suggested and merge it :)

skywhale:
Thank you @mkovaxx @bschwind for iterating on this and merging. I'm sorry for not being able to get back to it, but I'm so glad that it landed :)

PR #109, opened 2023-07-24
Add Bezier surface API

julianschuler:
This PR adds bezier surfaces to the high level API. While not using the same interface, this may be related to #47.

I also did a little deduplication and cleanup in the high level API, please let me know if I should factor this out to another PR.

### DIFF
--- a/crates/opencascade/src/primitives.rs
+++ b/crates/opencascade/src/primitives.rs
@@ -9,6 +9,7 @@ mod face;
 mod shape;
 mod shell;
 mod solid;
+mod surface;
 mod vertex;
 mod wire;
 
@@ -19,6 +20,7 @@ pub use face::*;
 pub use shape::*;
 pub use shell::*;
 pub use solid::*;
+pub use surface::*;
 pub use vertex::*;
 pub use wire::*;
 
--- a/crates/opencascade/src/primitives.rs
+++ b/crates/opencascade/src/primitives.rs
@@ -1,6 +1,6 @@
 use cxx::UniquePtr;
 use glam::DVec3;
-use opencascade_sys::ffi::{self};
+use opencascade_sys::ffi;
 
 mod boolean_shape;
 mod compound;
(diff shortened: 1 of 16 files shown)


bschwind:
Thanks for the PR! And sorry for the code mess and duplication, that's on my mental todo list 😅

It would be great if you could add a little example of building something with this API in `crates/opencascade/examples` (I should probably add a top-level directory for the examples to live in) so I can see how it's used, and others can see what kind of shapes can be created with it.

julianschuler:
No worries, I am really glad this project exists in the first place! Working with OpenCASCADE directly is really not fun and I am happy to contribute to a way more ergonomic API!

I now added an example where the bezier surface is used to create an airfoil like shape.

crates/opencascade/src/primitives.rs
```diff
@@ -298,6 +297,33 @@ impl Wire {
     pub fn freeform() {}
 }
 
+pub struct Surface {
+    inner: UniquePtr<ffi::HandleGeomSurface>,
+}
+
+impl Surface {
+    pub fn bezier(poles: &[&[DVec3]]) -> Self {
```
bschwind, line 305: We could make this API a bit more flexible:

```suggestion
    pub fn bezier<T: AsRef<[DVec3]>>(poles: impl IntoIterator<Item = T>) -> Self {
```

You'll then need to update various points in this function with calls to `.into_iter()` and `.as_ref()`

This should allow more flexibility in calling this function, allowing things like:

```rust
let surface = Surface::bezier([points1, points2]);
```

bschwind: Or maybe we can have a doubly nested `impl Iterator`? I'll look into that.

bschwind: @julianschuler ah, you can collect the iterator into a `Vec` for ease-of-use internally. It's more about presenting a nice API to end users and less about performance.

julianschuler: Thanks for the clarification, I now implemented it that way.

bschwind:
@julianschuler happy to merge this once the merge conflicts are resolved, sorry for the churn!

julianschuler:
That was quite a rebase, but now everything should be fixed.

bschwind (approved):
LGTM, another great PR. I liked the `from_make_x` refactors, that's a good cleanup.

Thanks for the contribution!

Issue #27, opened 2023-06-01
compile error


ltg1710:
thank for your open project

when i compile it (`cargo build`) in the envirment 
> cmake 3.23.0
> rustc 1.67
> windows 10 (home)
> visual studio 2022 community

The console give me some error message as fellow

``` text
cargo build
   Compiling opencascade-sys v0.1.0 (D:\opensource\OpenCascade\opencascade-rs\crates\opencascade-sys)
error: failed to run custom build command for `opencascade-sys v0.1.0 (D:\opensource\OpenCascade\opencascade-rs\crates\opencascade-sys)`

Caused by:
  process didn't exit successfully: `D:\opensource\OpenCascade\opencascade-rs\target\debug\build\opencascade-sys-8db3cae7a8db6976\build-script-build` (exit code: 101)
  --- stderr
  CMake Warning (dev) at adm/cmake/occt_toolkit.cmake:319 (if):
    Policy CMP0054 is not set: Only interpret if() arguments as variables or
    keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
    details.  Use the cmake_policy command to set the policy and suppress this
    warning.

    Quoted variables like "CSF_OpenGlLibs" will no longer be dereferenced when
    the policy is set to NEW.  Since the policy is not set the OLD behavior
    will be used.
  Call Stack (most recent call first):
    CMakeLists.txt:14 (include)
    src/TKernel/CMakeLists.txt:3 (OCCT_INCLUDE_CMAKE_FILE)
  This warning is for project developers.  Use -Wno-dev to suppress it.

  CMake Warning (dev) at adm/cmake/occt_toolkit.cmake:322 (if):
    Policy CMP0054 is not set: Only interpret if() arguments as variables or
    keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
    details.  Use the cmake_policy command to set the policy and suppress this
    warning.

    Quoted variables like "CSF_OpenGlesLibs" will no longer be dereferenced
    when the policy is set to NEW.  Since the policy is not set the OLD
    behavior will be used.
  Call Stack (most recent call first):
    CMakeLists.txt:14 (include)
    src/TKernel/CMakeLists.txt:3 (OCCT_INCLUDE_CMAKE_FILE)
  This warning is for project developers.  Use -Wno-dev to suppress it.

  CMake Warning (dev) at adm/cmake/occt_toolkit.cmake:319 (if):
[...]

PR #113, opened 2023-07-31
Fix the problem that the windows-gnu environment failed to find wrapper.hxx

ykong1337:
environment:
```
rustc: x86_64-pc-windows-gnu stable 1.71.0
linker: mingw-13.1.0-release-posix-seh-ucrt-rt_v11-rev1
```
error:
```
fatal error: opencascade-sys/include/wrapper.hxx: No such file or directory
  cargo:warning=    1 | #include "opencascade-sys/include/wrapper.hxx"
  cargo:warning=      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  cargo:warning=compilation terminated.
```
now, you have upgraded occt 7.7.1, it can be compiled normally in the windows environment, but still can not find ```wrapper.hxx```. #27 
i found out that i can manually set the environment variable:
```
CPLUS_INCLUDE_PATH = D:\workspace\opencascade-rs\crates
```
let it successfully find ```wrapper.hxx```(this problem may be because ```g++``` is a problem under different operating systems?)
then, I can let it find it directly in ```build.rs```, so as not to manually set the environment variable

### DIFF
--- a/crates/opencascade-sys/build.rs
+++ b/crates/opencascade-sys/build.rs
@@ -37,6 +37,9 @@ fn main() {
 
     if is_windows_gnu {
         build.define("OCC_CONVERT_SIGNALS", "TRUE");
+    }
+
+    if let "windows" = std::env::consts::OS {
         let current = std::env::current_dir().unwrap();
         build.include(current.parent().unwrap());
     } else {
--- a/crates/opencascade-sys/build.rs
+++ b/crates/opencascade-sys/build.rs
@@ -42,8 +42,6 @@ fn main() {
     if let "windows" = std::env::consts::OS {
         let current = std::env::current_dir().unwrap();
         build.include(current.parent().unwrap());
-    } else {
-        build.include("include");
     }
 
     build
@@ -51,6 +49,7 @@ fn main() {
         .flag_if_supported("-std=c++11")
         .define("_USE_MATH_DEFINES", "TRUE")
         .include(occt_include_path())
+        .include("include")
         .compile("wrapper");
 
     println!("cargo:rustc-link-lib=static=wrapper");
--- a/crates/opencascade-sys/build.rs
+++ b/crates/opencascade-sys/build.rs
@@ -37,6 +37,10 @@ fn main() {
 
     if is_windows_gnu {
         build.define("OCC_CONVERT_SIGNALS", "TRUE");
+        let current = std::env::current_dir().unwrap();
+        build.include(current.parent().unwrap());
+    } else {
+        build.include("include");
     }
 
     build
@@ -44,7 +48,6 @@ fn main() {
         .flag_if_supported("-std=c++11")
         .define("_USE_MATH_DEFINES", "TRUE")
         .include(occt_include_path())
-        .include("include")
         .compile("wrapper");
 
     println!("cargo:rustc-link-lib=static=wrapper");


ykong1337:
i did a few tests and probably found the reason (in windows system, native compilation):
i first use the official cxx.rs tutorial, the file name of the c++ library is ```lib.cc```, the name of the header file is ```lib.h```, and the wrapper is ```wrapper.h ```

build.rs:
```rust
let mut build = cxx_build::bridge("src/lib.rs");

     build
         .cpp(true)
         .flag_if_supported("-std=c++11")
         .include(format!("{}", dst. join("include").display()))
         .include("include")
         .compile("wrapper");
```
no problem at all

however, when i change the name of the c++ library to ```lib.cxx``` and ```lib.hxx```, and the wrapper is also ```wrapper.hxx```, it appears same error:

```
fatal error C1083: ???????: "example/include/wrapper.hxx": No such file or directory
```
So, i changed it again to:

build.rs
```rs
build.include(std::env::current_dir().unwrap().parent().unwrap())
```
it compiles successfully.

so i think that when the compiler searches for *.hxx header files, it needs to use the parent directory of the ```include``` directory.

bschwind:
That's... a bit weird, but wouldn't be the first time windows does things differently.

The change basically looks good from what you described, but let's try to find some definitive documentation on this behavior so we can leave it as a comment for future readers.

ykong1337:
sorry, i just found "official documentation for g++ importing extra include directories".
[Environment-Variables](https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html)

but I found some stackoverflow questions
[add-extra-include-lib-paths-to-mingw](https://stackoverflow.com/questions/4537377/add-extra-include-lib-paths-to-mingw)

combined with this problem, the ```-I``` parameter of g++ requires an absolute path in the windows environment.
so i modified the latest commit[ed5824b](https://github.com/bschwind/opencascade-rs/pull/113/commits/ed5824ba4cf5280effc6c3412eca4d36007e69ac) to ensure maximum compatibility in the windows environment.

eyinlijun:
@ykong1337 May I ask whether the wrapper was successfully generated on the WIN platform in the end?

ykong1337:
> @ykong1337 May I ask whether the wrapper was successfully generated on the WIN platform in the end?

yes, under ```gnu``` it generates ```libwrapper.a```

image

eyinlijun:
> > @ykong1337 May I ask whether the wrapper was successfully generated on the WIN platform in the end?
> 
> yes, under `gnu` it generates `libwrapper.a`
> 
> image

Does this mean that the library is temporarily unavailable in the MSVC toolchain in the WIN environment? If it is convenient, please add it to my wechatwechat(mobile:+86.158/0250/0611,Neil)

eyinlijun:
> > @ykong1337 May I ask whether the wrapper was successfully generated on the WIN platform in the end?
> 
> yes, under `gnu` it generates `libwrapper.a`
> 
> image

Does this mean that the library is temporarily unavailable in the MSVC toolchain in the WIN environment? If it is convenient, please add it to my wechatwechat(mobile:+86.158/0250/0611,Neil)

ykong1337:
> Does this mean that the library is temporarily unavailable in the MSVC toolchain in the WIN environment? If it is convenient, please add it to my wechatwechat(mobile:+86.158/0250/0611,Neil)

yes, you can see #27, my last reply, there are detailed error messages

DSchroer:
I tried this in a VM and can confirm it fixes the header error on MSVC.

crates/opencascade-sys/build.rs
```diff
@@ -39,6 +39,11 @@ fn main() {
         build.define("OCC_CONVERT_SIGNALS", "TRUE");
     }
 
+    if let "windows" = std::env::consts::OS {
```
bschwind, line 42: Sorry, last question:

Is there a reason we should use this instead of the `is_windows` variable? `is_windows` relates to the target, so I suppose we want to detect if windows is the _host_ OS, right?

bschwind:
@DSchroer thanks for testing!

crates/opencascade-sys/build.rs
ykong1337: Yes

bschwind: Got it, thanks! And sorry for the huge delay on merging this one.

PR #114, opened 2023-08-03
Move examples to their own lib and import them to the viewer bin

bschwind:


crates/viewer/src/main.rs
```diff
@@ -51,7 +46,8 @@ impl GameApp for ViewerApp {
     fn init(graphics_device: &mut GraphicsDevice) -> Self {
         // Model sourced from:
         // https://nist.gov/ctl/smart-connected-systems-division/smart-connected-manufacturing-systems-group/mbe-pmi-0
-        let keycap = Shape::read_step("crates/viewer/models/nist_ftc_06.step").unwrap();
+        // let keycap = Shape::read_step("crates/viewer/models/nist_ftc_06.step").unwrap();
+        let keycap = examples::gizmo::shape();
```
julianschuler, line 50: Currently the model for the viewer is always hardcoded, right? It might be a good idea to e.g. select the example via a command line argument, but this is probably out of the scope of this PR.
If you think this might be a good feature, I will try to implement this.

julianschuler (changes requested):
Just one minor remark/question, the rest LGTM!

crates/viewer/src/main.rs
julianschuler: I added an implementation for such a functionality in #117.

julianschuler (approved):
LGTM!

crates/viewer/src/main.rs
bschwind: Yep it was just hardcoded to get something displaying. Thanks for the PR, I'll take a look in a bit!

PR #117, opened 2023-08-05
Add command line argument to select example in viewer

julianschuler:
This small PR adds a command line argument to select the example displayed in the viewer application.

This depends on/is blocked by #114.

PR #118, opened 2023-08-05
Add command line argument to select example in viewer

julianschuler:
Supersedes #117 which was accidentally closed and can't be reopened due to a bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,12 +60,12 @@ The [higher level examples](./crates/opencascade/examples) use more ergonomic Ru
 
 ## Viewer Application
 
-There is currently an experimental viewer application based on WGPU, which will probably become the "main" way people use this crate. It currently visualizes a hardcoded model produced in Rust code, but will expand to be capable of loading Rust model code compiled to WASM, allowing faster compile times and more interactive inspection of the sketches and models.
+There is currently an experimental viewer application based on WGPU, which will probably become the "main" way people use this crate. It currently visualizes one of the examples, but will expand to be capable of loading Rust model code compiled to WASM, allowing faster compile times and more interactive inspection of the sketches and models.
 
-You can run the current viewer app with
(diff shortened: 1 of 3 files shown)


bschwind:
I was on the fence about this one, `clap` adds some extra compilation time (doesn't seem like much in practice), and it doesn't seem too scalable when increasing the number of examples.

But at the same time this is only temporary until I figure out how I want the WASM API to look and behave.

I thought of maybe adding all the examples into an array and letting you cycle through them with arrow keys, but changing between models would be a bit janky as they get built, and caching their built models might eat up a lot of memory.

Happy to merge it for now as it's an improvement over the existing code, just wanted to get your thoughts on it.

julianschuler:
I agree, this should definitely be considered as a temporary solution. Still, testing different examples currently requires rewriting the viewer code each time, so this improves the situation at least a little bit.

It may be interesting to have a look at the approach Fornjot takes, where each example is a package which itself calls the viewer. This requires more boilerplate for the examples, but does not require the viewer to be aware of them.

bschwind:
> It may be interesting to have a look at the approach Fornjot takes, where each example is a package which itself calls the viewer.

Does that mean the viewer would be a library crate, or is it literally doing the equivalent of `Command::new()` and calling the viewer?

That could be a nice intermediate approach too, but I'm fine with this current PR if we want to start with that first.

julianschuler:
It is a library crate (`fj-viewer`). I actually used it for displaying models made with OpenCascade in my current project by transforming the mesh representation, though I replaced it now with `three-d`. Still, it requires a main.rs which is essentially the same (though quite short) for all examples and would add a lot of boilerplate for an intermediate solution.

I think it is best to keep this as simple as possible for now, so lets merge this.

bschwind (approved):
LGTM!

Issue #22, opened 2023-03-22
Feature wish-list


bschwind:
In no particular order

- [ ] Create part 3-views, with dimension annotations
- [ ] Automatic fillets at the intersection of joined solids
- [ ] 3D viewer with material/texture support
- [ ] Lightweight model code (written in Rust) which compiles to WASM for the 3D viewer to interpret as commands and build the model
- [ ] Ability to define GUI widgets (egui?) on part features, for adjustability in the viewer app
- [ ] Ability to import Kicad PCB files directly (no intermediate exports from kicad needed)
- [ ] Ability to import STEP files and reference parts,
- [ ] Ability to import SVG files as sketches for later use
- [ ] Projection of parts onto planes or surfaces
- [ ] STEP export
- [ ] Ability to "tag" certain features as you build a part, to reference later when building (example: drill a hole with a cylinder, but tag the bottom face of the cylinder to have a location to start from for future operations
- [ ] Some sort of "timeline" in the viewer app to scrub through all the steps and see the model get created at each step
- [ ] Having a "source map" of objects to the lines of code which created it would be super cool, though maybe infeasible
- [ ] Ability to create a quick sketch in the viewer app and export that sketch to Rust code. Essentially, bidirectional editing between code and GUI would be ideal
- [ ] Font shape support for putting text on shapes

julianschuler:
Regarding the renderer: It might be worth having a look at [`three-d`](https://docs.rs/three-d/0.15.0/three_d/) since it already provides a lot of features like lighting, physically-based rendering, shadow maps, orbital control, etc.

It is quite simple to transfer the mesh representation of this library to the one `three-d` uses, see [here](https://github.com/julianschuler/concavum-customizer/blob/21c780c24c04a1d732f13842b8858e6c8f67237d/src/viewer/model.rs#L25) for a proof of concept.

PR #121, opened 2023-08-06
Make tolerance of meshing operations configurable

julianschuler:
This PR makes the triangulation tolerance used in the meshing operations configurable. It also introduces error handling for the meshing operations as well as doing some deduplication of the AdhocShape/Solid methods with the shape ones.

### DIFF
--- a/crates/opencascade/src/mesh.rs
+++ b/crates/opencascade/src/mesh.rs
@@ -12,12 +12,12 @@ pub struct Mesh {
 }
 
 pub struct Mesher {
-    inner: UniquePtr<ffi::BRepMesh_IncrementalMesh>,
+    pub(crate) inner: UniquePtr<ffi::BRepMesh_IncrementalMesh>,
 }
 
 impl Mesher {
-    pub fn new(shape: &Shape) -> Self {
-        let inner = ffi::BRepMesh_IncrementalMesh_ctor(&shape.inner, 0.01);
+    pub fn new(shape: &Shape, triangulation_tolerance: f64) -> Self {
+        let inner = ffi::BRepMesh_IncrementalMesh_ctor(&shape.inner, triangulation_tolerance);
 
         if !inner.IsDone() {
             // TODO(bschwind) - Add proper Error type and return Result.
--- a/crates/opencascade/src/mesh.rs
+++ b/crates/opencascade/src/mesh.rs
@@ -22,9 +22,10 @@ impl Mesher {
     pub fn try_new(shape: &Shape, triangulation_tolerance: f64) -> Result<Self, Error> {
         let inner = ffi::BRepMesh_IncrementalMesh_ctor(&shape.inner, triangulation_tolerance);
 
-        match inner.IsDone() {
-            true => Ok(Self { inner }),
-            false => Err(Error::TriangulationFailed),
+        if inner.IsDone() {
+            Ok(Self { inner })
+        } else {
+            Err(Error::TriangulationFailed)
         }
     }
 
--- a/crates/opencascade/src/mesh.rs
+++ b/crates/opencascade/src/mesh.rs
@@ -35,8 +35,7 @@ impl Mesher {
         let mut normals = vec![];
         let mut indices = vec![];
 
-        let triangulated_shape = ffi::TopoDS_Shape_to_owned(self.inner.pin_mut().Shape());
-        let triangulated_shape = Shape { inner: triangulated_shape };
+        let triangulated_shape = Shape::from_shape(self.inner.pin_mut().Shape());
 
         for face in triangulated_shape.faces() {
             let mut location = ffi::TopLoc_Location_ctor();
--- a/crates/opencascade/src/mesh.rs
+++ b/crates/opencascade/src/mesh.rs
@@ -1,4 +1,7 @@
-use crate::primitives::{FaceOrientation, Shape};
+use crate::{
+    primitives::{FaceOrientation, Shape},
+    Error,
+};
 use cxx::UniquePtr;
 use glam::{dvec2, dvec3, DVec2, DVec3};
 use opencascade_sys::ffi;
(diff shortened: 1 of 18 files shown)


bschwind:
Nice! You probably already know but the goal is to completely remove the AdHocShape stuff. As the name implies that was just what I threw together to get my keyboard case built. Thanks for starting the de-duplication!

julianschuler:
Oh, I actually didn't know that. Is there a roadmap or issue somewhere where this kind of general direction is documented? I am happy to help and would like to ensure that my contributions to the API align with your ideas for it!

bschwind:
I don't have anything documented on AdHocShape as that appeared unintentionally, but the general high level features I want are documented here:

https://github.com/bschwind/opencascade-rs/issues/22

bschwind:
@julianschuler just FYI, I've been traveling the past week or two so I haven't had much time to review things. I'll be back to a more regular schedule next week :)

crates/opencascade/src/mesh.rs
```diff
@@ -12,22 +15,20 @@ pub struct Mesh {
...
-        let inner = ffi::BRepMesh_IncrementalMesh_ctor(&shape.inner, 0.01);
+    pub fn try_new(shape: &Shape, triangulation_tolerance: f64) -> Result<Self, Error> {
+        let inner = ffi::BRepMesh_IncrementalMesh_ctor(&shape.inner, triangulation_tolerance);
 
-        if !inner.IsDone() {
-            // TODO(bschwind) - Add proper Error type and return Result.
-            panic!("Call to ffi::BRepMesh_IncrementalMesh_ctor failed");
+        match inner.IsDone() {
+            true => Ok(Self { inner }),
+            false => Err(Error::TriangulationFailed),
```
bschwind, line 27: I have a slight preference for `if` statements when it's a simple true/false match but this is a minor detail.

crates/viewer/src/main.rs
```diff
@@ -76,7 +76,7 @@ impl GameApp for ViewerApp {
         let example = Example::parse();
         let shape = example.shape();
 
-        let mesh = shape.mesh();
+        let mesh = shape.mesh().unwrap();
```
bschwind, line 79: Let's use an `expect()` here:

```suggestion
        let mesh = shape.mesh().expect("example shape should yield a valid triangulation");
```

examples/src/high_level_bottle.rs
```diff
@@ -34,7 +34,7 @@ pub fn shape() -> Shape {
     let neck_radius = thickness / 4.0;
     let neck_height = height / 10.0;
 
-    let mut neck = AdHocShape::make_cylinder(dvec3(0.0, 0.0, height), neck_radius, neck_height);
+    let neck = AdHocShape::make_cylinder(dvec3(0.0, 0.0, height), neck_radius, neck_height);
     neck.union(&body);
```
bschwind, line 38: This change is now incorrect, as `union()` is no longer a mutable function, but something that returns a new `BooleanShape`.

`neck.union(&body);` now needs to be `let neck = neck.union(&body);`

My suggestion would be to add a `#[must_use]` attribute to all functions which return `BooleanShape` so it's harder to make this mistake.

bschwind: Perhaps we'd also want to add in-place, mutable versions of the boolean operations (`union_mut()` or something like that).

bschwind (changes requested):
Looking good but the `high_level_bottle` and `keyboard_case` examples need to be fixed as they are now throwing away the results of some boolean shape operations.

examples/src/high_level_bottle.rs
julianschuler: Good point, I think this is an artifact of a previous PR.

I actually prefer returning a new object instead of the mutable option as it allows for chaining operations. However, this doesn't cover cases like iterating over a some shapes and union it with another one (though this can be done with fold(), but is not as straight forward). What do you think about something like a `union_shapes()` function which takes an iterator of objects implementing `Into<Shape>`?

On another note, some operations like `fillet()` mutate the object while others like `union()` return a new one. It may be a good idea to select one of the styles to support to make the coding experience more uniform:

Example when using exclusively the first style:
```rust
let outer_box = case_outer_box()
    .subtract(&inner_box)
    .fillet(1.3)
    .union(&top_shelf)
    .union(&bottom_shelf)
    .subtract(&usb_cutout)
    .union_shapes(&SUPPORT_POSTS)
```

Example when using exclusively the second one:
```rust
let mut outer_box = case_outer_box();

outer_box.subtract(&inner_box);
outer_box.fillet(1.3);

outer_box.union(&top_shelf);
outer_box.union(&bottom_shelf);
outer_box.subtract(&usb_cutout);

for support_post in SUPPORT_POSTS {
    outer_box.union(&support_post.shape());
}
```

Which one do you prefer or should we rather keep the mixed approach?

crates/opencascade/src/mesh.rs
julianschuler: Done.

crates/viewer/src/main.rs
julianschuler: Good point, fixed.

examples/src/high_level_bottle.rs
bschwind: I think the first example is nicer, though it would be nice to also support the second style if we can come up with an API that isn't ugly. My only first thought is something like `subtract_mut` and `fillet_mut` but that's still pretty ugly.

Probably for now it's better to support a chainable API, and use the `#[must_use]` attribute so users don't make the mistake of ignoring the result.

julianschuler: This was quite something, but I now made almost the entire API chainable and added `#[must_use]` whenever `&self` is used and a new shape object is returned. I also adjusted the examples accordingly and IMO their code became a lot nicer in a lot of places!

Something like `union_shapes()` and the `*_mut` operations should probably best be done in another PR.

bschwind: Very nice! I'll try to get a more thorough review on this soon.

> should probably best be done in another PR

Agreed, this one has blown up a fair amount from my suggestions 😅

crates/opencascade/src/primitives/face.rs
```diff
@@ -18,20 +18,25 @@ impl AsRef<Face> for Face {
...
         let inner = ffi::TopoDS_Face_to_owned(face);
 
         Self { inner }
     }
 
+    fn from_make_face(make_face: UniquePtr<ffi::BRepBuilderAPI_MakeFace>) -> Self {
+        Self::from_face(make_face.Face())
+    }
+
+    #[must_use]
```
bschwind, line 31: This is fine to have but we can probably relax the usage of `#[must_use]` for constructors, since callers don't have an instance of the struct to mutate.

Basically any method with `&self` which sounds like it might mutate `self`, but actually returns a new `Shape` will benefit from having `#[must_use]`.

bschwind (approved):
Looks great! Just one minor nitpick with the `#[must_use]` attribute but I've tested this and the code looks good to merge.

Thanks for the big improvement!

crates/opencascade/src/primitives/face.rs
bschwind: @julianschuler I pushed a few changes to address this comment. Once it passes CI I'll go ahead and merge it :)

PR #133, opened 2023-09-04
Support loading a STEP file from the viewer app args

### LINKED ISSUE #132: Expand viewer CLI args parsing
It would be nice to allow the viewer to specify:

* A STEP file to read and display

or

* An STL file to read and display

or

* The name of an example to run

bschwind:
Closes #132

TODO

- [ ] Support STL files

### DIFF
--- a/crates/viewer/src/main.rs
+++ b/crates/viewer/src/main.rs
@@ -2,9 +2,9 @@ use crate::{
     edge_drawer::{EdgeDrawer, LineBuilder, LineVertex3, RenderedLine},
     surface_drawer::{CadMesh, SurfaceDrawer},
 };
-use clap::Parser;
+use clap::{Parser, ValueEnum};
 use glam::{vec3, DVec3, Mat4};
-use opencascade::primitives::Shape;
+use opencascade::{adhoc::AdHocShape, primitives::Shape};
 use simple_game::{
     graphics::{
         text::{AxisAlign, StyledText, TextAlignment, TextSystem},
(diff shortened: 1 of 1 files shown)


crates/viewer/src/main.rs
```diff
@@ -75,8 +85,16 @@ impl GameApp for ViewerApp {
...
-        let example = Example::parse();
-        let shape = example.shape();
+        let args = AppArgs::parse();
+
+        let shape = if let Some(step_file) = args.step_file {
+            Shape::read_step(step_file).expect("Failed to read STEP file, {step_file}")
+        } else if let Some(example) = args.example {
+            example.shape()
+        } else {
+            println!("Warning - no example or STEP file specified, you get a default cube.");
```
julianschuler, line 95: Consider using eprintln! instead to write to stderr.

```suggestion
            eprintln!("Warning - no example or STEP file specified, you get a default cube.");
```

julianschuler (changes requested):
I really like the idea of this PR! I have just one minor remark.

crates/viewer/src/main.rs
julianschuler: On another note: I am unable to resolve this conversation now (the resolve button is simply missing), but if you review any PRs, it is always present. Do you have an idea why this might by the case?

bschwind: I think it automatically marks it as resolved when a suggestion is committed to the branch from the github comment UI.

PR #134, opened 2023-09-08
Add creating wire from ordered points

julianschuler:
This PR adds the option to create a wire from a list of ordered points. It also uses this new function for simplifying `extrude_polygon()` in the ad-hoc API.

### DIFF
--- a/crates/opencascade/src/primitives/wire.rs
+++ b/crates/opencascade/src/primitives/wire.rs
@@ -45,7 +45,8 @@ impl Wire {
         Self::from_wire(make_wire.pin_mut().Wire())
     }
 
-    pub fn from_ordered_points(points: &[DVec3]) -> Self {
+    pub fn from_ordered_points(points: impl IntoIterator<Item = DVec3>) -> Self {
+        let points: Vec<_> = points.into_iter().collect();
         let mut make_wire = ffi::BRepBuilderAPI_MakeWire_ctor();
 
         if let (Some(first), Some(last)) = (points.first(), points.last()) {
--- a/crates/opencascade/src/primitives/wire.rs
+++ b/crates/opencascade/src/primitives/wire.rs
@@ -3,6 +3,7 @@ use std::iter::once;
 use crate::{
     angle::{Angle, ToAngle},
     primitives::{make_dir, make_point, make_vec, Edge, Face, Shape},
+    Error,
 };
 use cxx::UniquePtr;
 use glam::{dvec3, DVec3};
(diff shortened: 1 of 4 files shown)


crates/opencascade/src/primitives/wire.rs
```diff
@@ -43,6 +45,19 @@ impl Wire {
         Self::from_wire(make_wire.pin_mut().Wire())
     }
 
+    pub fn from_ordered_points(points: &[DVec3]) -> Self {
+        let mut make_wire = ffi::BRepBuilderAPI_MakeWire_ctor();
+
+        if let (Some(first), Some(last)) = (points.first(), points.last()) {
+            for window in points.windows(2).chain(once([*last, *first].as_slice())) {
```
bschwind, line 52: If we passed in just two points, would this create a "degenerate" wire which overlaps itself exactly? The code as it is here is very elegant, but I'm not sure if it handles that case of "two points" correctly.

crates/opencascade/src/primitives/wire.rs
```diff
@@ -43,6 +45,19 @@ impl Wire {
         Self::from_wire(make_wire.pin_mut().Wire())
     }
 
+    pub fn from_ordered_points(points: &[DVec3]) -> Self {
```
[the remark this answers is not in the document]
julianschuler: Good point, thanks! I changed this now accordingly.

crates/opencascade/src/primitives/wire.rs
julianschuler: I tested this now a little bit and it seams to work fine with 2 different points, though it and any resulting face will result in such a "degenerate" case. Trying to extrude such a face will result an a non-triangularizable shape, so it might be easiest be to restrict this function to at least three points.

julianschuler: The function now returns an error if less than three points were given.

crates/opencascade/src/primitives/wire.rs
```diff
@@ -43,6 +46,23 @@ impl Wire {
         Self::from_wire(make_wire.pin_mut().Wire())
     }
 
+    pub fn from_ordered_points(points: impl IntoIterator<Item = DVec3>) -> Result<Self, Error> {
+        let points: Vec<_> = points.into_iter().collect();
+        let mut make_wire = ffi::BRepBuilderAPI_MakeWire_ctor();
+
+        if points.len() < 3 {
```
bschwind, line 53: I think we can still accept 2 points - just create a segment and return it directly.

We can return `Error::NotEnoughPoints` for 1 or fewer points, though.

julianschuler: Oh, I wrongly assumed that wires are closed sequence of edges. I now adjusted the implementation to allow using 2 points as well.

bschwind (approved):
Looks good, thanks for the change!

PR #142, opened 2023-10-16
add Shell::loft(wires) and underlying machinery

mkovaxx:
Adds `Shell::loft(wires)` and underlying machinery.
Essentially copy-paste of `Solid::loft(wires)` with a tiny change.


bschwind (approved):
LGTM! We _could_ move the duplicated logic to `shape.rs` and then call it from `Solid::loft` and `Shell::loft` but it's so trivial I don't think it's worth it.

mkovaxx:
> LGTM! We _could_ move the duplicated logic to `shape.rs` and then call it from `Solid::loft` and `Shell::loft` but it's so trivial I don't think it's worth it.

Agreed. It seems to me like it's not possible to avoid having two separate functions, because we want the return type to be specific.

PR #145, opened 2023-10-18
Offset for Face and Wire

### LINKED ISSUE #144: Offset for Face and Wire
Splitting this off from https://github.com/bschwind/opencascade-rs/issues/72 because [`Shape::offset_surface()`](https://github.com/bschwind/opencascade-rs/blob/b6c7a05f130d297106df9b2d54eed60f196a2fde/crates/opencascade/src/primitives/shape.rs#L355) is already implemented.

OCC has [BRepOffsetAPI_MakeOffset](https://dev.opencascade.org/doc/refman/html/class_b_rep_offset_a_p_i___make_offset.html) for this.
Can work with either a Face or a Wire.
Supports 3 join types: arc, tangent, and intersection.
Has a "ToApprox" boolean setting to use only linear segments and circular arcs in the output.

mkovaxx:
Closes #144.


bschwind (commented):
Nice! Let's do a few things to finish this up:

* Format the code with `cargo +nightly fmt`
* Fix the compilation error
* Add a usage of this feature, either in a new example or add it to an existing example if it makes sense

mkovaxx:
I need to figure out a couple of things, so this is still a draft. Stay tuned! :)

bschwind (commented):
All the binding code looks good! I know I was sitting with you when you created the example file, but I think we should add a bit more to it to demonstrate what's going on:

Here's an example of a chevron shape with acute and obtuse angles:

```rust
let solid_1 = Workplane::xy()
    .translated(DVec3::new(32.0 * i as f64, 0.0, 0.0))
    .sketch()
    .move_to(0.0, 10.0)
    .line_to(5.0, 5.0)
    .line_to(5.0, -5.0)
    .line_to(0.0, 0.0)
    .line_to(-5.0, -5.0)
    .line_to(-5.0, 5.0)
    .close()
    .offset(1.0, join_type) // offset a wire
    .to_face()
    .extrude(DVec3::new(0.0, 0.0, 8.0))
    .into_shape();
```

I think if we add that shape along with positive and negative offsets, it will more readily demonstrate what the feature does.

While testing this I also had crashes with positive offsets for the `Tangent` join type and I'm not sure why.

It's nice to demonstrate that you can offset a face _or_ a wire, but ultimately they result in the same exact shape so the visuals don't really show you what's happening so much.

mkovaxx:
Uh-oh I'm getting the segfaults too...

bschwind:
@mkovaxx that particular join type just might not be supported yet, I'd have to look at the code to be sure.

mkovaxx:
@bschwind I propose that we put `JoinType::Tangent` out of scope for now, and try to add it later.
Issue to track that work: #153

bschwind:
> I propose that we put JoinType::Tangent out of scope for now, and try to add it later.

Makes sense! Shall we mark this ready for review?

bschwind (approved):
LGTM, thanks!

PR #147, opened 2023-11-08
Extend Shape::faces_along_ray to return a LineFaceHitPoint with more info

### LINKED ISSUE #146: Enhance faces_along_ray() function
Currently the function just returns a Face and a point on the face where it intersected. It will be far more useful to also return:

* The parameter for the line at the point it intersects the face
* The U and V parameters where the intersection occurs on the face
* The surface normal at the intersection point

This should probably just be packaged up in a struct now instead of a tuple, maybe called `IntersectionPoint` or something like that which contains all of the above intersection information.

mkovaxx:
Closes #146

As Mike Tyson said (paraphrasing):
> Everybody has a plan until they get hit in the face.

By a line. With parameters, and stuff.

### DIFF
--- a/crates/opencascade/src/primitives/shape.rs
+++ b/crates/opencascade/src/primitives/shape.rs
@@ -542,7 +542,7 @@ impl Shape {
     }
 
     // TODO(bschwind) - Convert the return type to an iterator.
-    pub fn faces_along_ray(&self, ray_start: DVec3, ray_dir: DVec3) -> Vec<(Face, DVec3)> {
+    pub fn faces_along_ray(&self, ray_start: DVec3, ray_dir: DVec3) -> Vec<LineFaceHitPoint> {
         let mut intersector = ffi::BRepIntCurveSurface_Inter_ctor();
         let tolerance = 0.0001;
         intersector.pin_mut().Init(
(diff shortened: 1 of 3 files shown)


crates/opencascade/src/primitives/shape.rs
```diff
@@ -542,7 +542,7 @@ impl Shape {
     }
 
     // TODO(bschwind) - Convert the return type to an iterator.
-    pub fn faces_along_ray(&self, ray_start: DVec3, ray_dir: DVec3) -> Vec<(Face, DVec3)> {
+    pub fn faces_along_ray(&self, ray_start: DVec3, ray_dir: DVec3) -> Vec<LineFaceHitPoint> {
```
bschwind, line 545: Let's go ahead and also rename this function to `faces_along_line` since that's technically what it's doing.

examples/src/keycap.rs
```diff
@@ -210,16 +210,12 @@ pub fn shape() -> Shape {
...
-
-            a_dist.total_cmp(&b_dist)
-        });
-
-        let (face_target, _) = faces.get(0).expect("We should have a face to extrude to");
-        let post = circle.extrude_to_face(&keycap, face_target);
+        let faces = keycap.faces_along_ray(origin, workplane.normal());
+        let nearest_hit = faces
+            .iter()
+            .min_by(|hit_a, hit_b| hit_a.t.partial_cmp(&hit_b.t).unwrap())
```
bschwind, line 216: A little out of scope for this PR but it would be nice to have dedicated iterators with convenience functions for this sort of thing. Take a look at `FaceIterator` for what I mean, we could have things like `farthest`, `farthest_negative`, `nearest`, etc.

bschwind (approved):
Looks good, just needs a rename!

examples/src/keycap.rs
mkovaxx: I've added an issue to track this idea:
https://github.com/bschwind/opencascade-rs/issues/149

bschwind (approved):
LGTM!

PR #155, opened 2023-11-26
Sweep Faces and Wires along a path

### LINKED ISSUE #43: Support pipes
Reference: https://dev.opencascade.org/doc/overview/html/occt_user_guides__modeling_algos.html#occt_modalg_7_4

mkovaxx:
Closes #43.

@bschwind Both face and wire sweeps work as expected using the simpler MakePipe API.
However, taking the `union` of a `Solid` and a `Shell` appears to be broken, so I split the example into two.


bschwind (approved):
One little nit-pick, but this looks great! I ran both examples, everything seems to work well.

PR #158, opened 2024-01-10
rerun build script if env var changes

DSchroer:
Rerun the `opencascade-sys` build script when `DEP_OCCT_ROOT` changes. Removing the need to run `cargo clean` every time that value is modified.


bschwind (approved):
LGTM!

PR #159, opened 2024-01-12
fix: double free segfault when using Handle_Poly_Triangulation_ctor

DSchroer:
Handle_Poly_Triangulation_ctor caused a segfault since the underlying Triangulation would be freed twice. Once when the new handle was dropped and once when the UniquePtr was dropped. By passing ownership to the triangulation this is fixed.


bschwind (approved):
LGTM, thanks as always :)

Random aside - do you think `unique_ptr` was the right choice for all of the bridge code, or would `shared_ptr` have been the better approach? This is one aspect of the bindings I'm still unsure about.

PR #161, opened 2024-01-12
fix: emscripten unable to find OCCT

DSchroer:
When compiling for WASM. The new build system fails to find OCCT. This is due to a bug in emscripten and there is a work around.

https://github.com/emscripten-core/emscripten/issues/19243


bschwind (approved):
Very cool! How easy is it overall to compile this library to wasm? That's something I want to do eventually (running stuff in the browser) but haven't gotten into it at all yet.

Issue #160, opened 2024-01-12
Cross compilation broken for occt-sys


DSchroer:
If you try to cross compile using the builtin occt-sys. It will always compile OCCT for the HOST architecture rather than the TARGET architecture. 

This is due to it being a build-dependency.
```
opencascade-sys: built for TARGET
    build.rs: built/run for HOST with env TARGET=TARGET
        occt-sys: built for HOST (since its a dep of build.rs)
            build.rs: build for HOST with env TARGET=HOST
```

bschwind:
This should be fixed now, let me know if it's not!

PR #163, opened 2024-01-16
fix: allow cross compilation under builtin flag

DSchroer:
Fix for https://github.com/bschwind/opencascade-rs/issues/160

`occt-sys` no longer builds OCCT in the build.rs file. This allows it to be re-included in the workspace and the build still only triggered when the `builtin` feature is enabled. 

This also now passes the correct ENV variables to the build so it is built for TARGET and not HOST. 

Finally added CI to test these builds work on various platforms. 

Had to rename `Handle_*` to `Handle*` for MSVC.

### DIFF
--- a/crates/occt-sys/Cargo.toml
+++ b/crates/occt-sys/Cargo.toml
@@ -3,7 +3,7 @@ name = "occt-sys"
 description = "Static build of the C++ OpenCascade CAD Kernel for use as a Rust dependency"
 authors = ["Brian Schwind <brianmschwind@gmail.com>", "Matěj Laitl <matej@laitl.cz>"]
 license = "LGPL-2.1"
-version = "0.3.0"
+version = "0.4.0"
 edition = "2021"
 repository = "https://github.com/bschwind/opencascade-rs"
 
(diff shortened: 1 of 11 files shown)


crates/occt-sys/Cargo.toml
```diff
@@ -148,7 +148,7 @@ exclude = [
     # "OCCT/src/BRepCheck/*",
     # "OCCT/src/MAT2d/*",
     "OCCT/src/ExpToCasExe/*",
-    # "OCCT/src/STEPControl/*",
+    "OCCT/src/STEPControl/*",
```
bschwind, line 151: This might have implications on whether or not I can publish to crates.io, we're currently right at the 10MB crate limit. We'll see how it goes

DSchroer: I had issues where the STEPControl.hxx headers were missing on windows when it was not included.

bschwind:
@DSchroer I'd be interested in revisiting this - since this PR was opened I have requested a larger crate size for `occt-sys` which means I don't need to do as many `exclude` shenanigans in `Cargo.toml`, as well as less patching of the occt src directory.

fidoriel:
I am interested in a working cross compile chain. Any chance to  finish this project @DSchroer? If not, I am going to try it ;)

bschwind:
@fidoriel please do! I'll do my best to help review/test if any PRs are posted.

fidoriel:
To sum up:
the crates upload limit is overcome. Nevertheless, the windows thing deserves its own PR, being out of scope for this one. Hence I am not on windows, so cannot (and will not) test. And therefor extract the cross compile part.

So, a proper cleanup and squash is needed? I use currently a fork with this patch applied ;)

bschwind:
@fidoriel yep that sounds about right!

bschwind:
I think this has now been superseded by #185 and #190. @DSchroer sorry this lingered so long, but please feel free to open another PR if there's more you want to add.

I finally have a working Windows machine so I'll also work to get everything working on that platform (hopefully) soon too.

PR #164, opened 2024-01-19
Add copy mingw runtime to the output directory

yk0n9:
The binary generated from cross-compiling from linux to windows-gnu cannot run because it lacks the [mingw runtime](https://stackoverflow.com/questions/46728353/mingw-w64-whats-the-purpose-of-libgcc-s-seh-dll) (mingw-w64 under the linux platform).

So I improved this build script so that it copies mingw's runtime to the output directory. In this way, we distribute binaries and these dynamic link libraries, and users do not need the mingw environment.

mingw-dll

But we need to wait for #163, the current occt cross-compiled product is still host, not target.

### DIFF
--- a/crates/opencascade-sys/build.rs
+++ b/crates/opencascade-sys/build.rs
@@ -51,6 +51,29 @@ fn main() {
 
     if is_windows_gnu {
         build.define("OCC_CONVERT_SIGNALS", "TRUE");
+        let out_dir = std::env::var("OUT_DIR").expect("No OUT_DIR environment variable defined");
+        let out_dir = std::path::Path::new(&out_dir).join("../../..");
+        let mut dll_paths = vec![];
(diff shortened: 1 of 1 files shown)


bschwind:
So I finally got a working Windows machine, and was able to get this project compiling on the "vanilla" Windows Rust experience (`stable-x86_64-pc-windows-msvc`) with #196

Looking more closely at this PR, I'm not sure it should be the responsibility of this repo to package mingw DLLs in the output. If you're packaging a particular application which uses this crate and cross-compiles from linux to windows-gnu, I would add the DLL packaging in _that_ application as opposed to this one.

I see potential for this list of DLLs to get outdated in the future as well, which is another reason I hesitate to include it here.

Let me know if I'm misunderstanding, though.

PR #123, opened 2023-08-17
Initial WASM API

bschwind:
Closes #67

PR #173, opened 2024-06-02
First pass at using wit-bindgen/wasmtime and WIT interfaces

bschwind:
Starts #67

Supersedes #123

TODO

- [x] Fix all clippy warnings/errors
- [x] Use the `wit-component` crate to [convert the wasm bundle to a wasm component](https://github.com/bytecodealliance/wasmtime/blob/28ee78b654e45cfa5837857d7880d8c40d571cb4/examples/component/main.rs#L32-L37)
- [x] Have live-reloading WASM files
- [x] Log model errors in the viewer app instead of crashing
- [x] Migrate some more modeling API functions to at least be able to create some interesting solids and shapes
- [ ] (nice to have) - Figure out auto-formatting for WIT files

### DIFF
--- a/crates/model-api/src/workplane.rs
+++ b/crates/model-api/src/workplane.rs
@@ -1,0 +1,204 @@
+use crate::{
+    angle::{Angle, RVec},
+    primitives::{Edge, Wire},
+};
+use glam::{dvec3, DAffine3, DMat3, DVec3, EulerRot};
+
+#[derive(Debug, Copy, Clone)]
+pub enum Plane {
+    XY,
+    YZ,
+    ZX,
+    XZ,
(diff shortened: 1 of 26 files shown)


crates/model-api/src/workplane.rs
```diff
@@ -0,0 +1,204 @@
...
+    XZ,
+    YX,
+    ZY,
+    Front,
+    Back,
+    Left,
+    Right,
+    Top,
+    Bottom,
+    Custom { x_dir: (f64, f64, f64), normal_dir: (f64, f64, f64) },
```
mkovaxx, line 21: Should a custom workplane have an origin too?

PR #175, opened 2024-06-06
feat: variable radius pipe

mkovaxx:
- add `Wire::sweep_along_with_radius_values()` to produce a `Shell`
- add `Face::sweep_along_with_radius_values()` to produce a `Solid`


crates/opencascade/src/primitives/face.rs
```diff
@@ -213,6 +214,41 @@ impl Face {
...
+        // let law_function = ffi::Law_Interpol_into_Law_Function(interpol);
+        // let law_handle = ffi::Law_Function_to_handle(law_function);
+
+        let profile_shape = ffi::cast_face_to_shape(&self.inner);
+        let mut make_pipe_shell = ffi::BRepOffsetAPI_MakePipeShell_ctor(&path.inner);
+        let with_contact = false;
+        let with_correction = true;
+        // make_pipe_shell.pin_mut().SetLaw(profile_shape, &law_handle, with_contact, with_correction);
+        println!("1 POTATO");
+        make_pipe_shell.pin_mut().Add(profile_shape, with_contact, with_correction);
```
mkovaxx, line 243: TODO: figure out why it segfaults here 😅

mkovaxx: Done.

mkovaxx:
@bschwind Here's the dilemma I'd appreciate your input on.

From the user's perspective, I think it makes more sense that sweeping a `Wire` profile yields a `Shell`, and sweeping a `Face` profile yields a `Solid`.

However, the underlying [`BRepOffsetAPI_MakePipeShell`](https://dev.opencascade.org/doc/refman/html/class_b_rep_offset_a_p_i___make_pipe_shell.html) machinery always takes a `Wire` for the profile. To produce a `Solid` instead of a `Shell`, it has a `MakeSolid()` method.

Thoughts?

bschwind:
> However, the underlying [BRepOffsetAPI_MakePipeShell](https://dev.opencascade.org/doc/refman/html/class_b_rep_offset_a_p_i___make_pipe_shell.html) machinery always takes a Wire for the profile. To produce a Solid instead of a Shell, it has a MakeSolid() method.

Could we do something like this?

* When the user sweeps a wire, they get a shell
* When the user sweeps a face, convert it to a wire (using [outer_wire](https://github.com/bschwind/opencascade-rs/blob/e67ddf01ed218b257cbe8b022c8432e1779a6e86/crates/opencascade/src/primitives/wire.rs#L166)), sweep it, then call `MakeSolid()`

mkovaxx:
@bschwind One step closer!

Now it fails with:
```
libc++abi: terminating due to uncaught exception of type StdFail_NotDone
```

bschwind:
> Now it fails with:
> ```
> libc++abi: terminating due to uncaught exception of type StdFail_NotDone
> ```

This happens with fallible APIs such as a fillet or chamfer with a size which is just too large for the mode. I'd start with a simpler usage of the API and see if the error still occurs.

crates/opencascade/src/primitives/face.rs
```diff
@@ -213,6 +214,42 @@ impl Face {
...
+
+        for (index, (t, radius)) in radius_values.into_iter().enumerate() {
+            array.pin_mut().SetValue(index as i32 + 1, &make_point2d(dvec2(t, radius)));
+        }
+
+        let mut interpol = ffi::Law_Interpol_ctor();
+        let is_periodic = false;
+        interpol.pin_mut().Set(&array, is_periodic);
+        let law_function = ffi::Law_Interpol_into_Law_Function(interpol);
+        let law_handle = ffi::Law_Function_to_handle(law_function);
```
bschwind, line 235: I'm usually against premature de-duplication, but this code is complicated enough that we should probably extract parts to some functions.

* One could take a `radius_values: impl IntoIterator<Item = (f64, f64)>,` and return a `UniquePtr<HandleLawFunction>`
* Another that takes a profile `Wire` and returns a `UniquePtr<BRepOffsetAPI_MakePipeShell>`.

The main difference in code paths is one calls `make_solid()` and the other doesn't.

bschwind (commented):
Looking good! I think there's one place where we should deduplicate some code but otherwise it's all working well. You can make some pretty cool shapes with this.

crates/opencascade/src/law_function.rs
```diff
@@ -0,0 +1,22 @@
+use cxx::UniquePtr;
+use glam::dvec2;
+use opencascade_sys::ffi;
+
+use crate::primitives::make_point2d;
+
+#[must_use]
+pub(crate) fn law_function_from_graph(
```
mkovaxx, line 8: I thought that exposing `LawFunction` on the public API would be premature at this point.

crates/opencascade/src/primitives/face.rs
mkovaxx: Done! :)

crates/opencascade/src/law_function.rs
bschwind: Agreed, let's keep it local to the crate for now.

PR #181, opened 2024-09-01
feat: gtransform API

DSchroer:
Adding support for the GTransform API which lets you manipulate and apply transform matrices directly. Unlocking a way to interoperate between Rust Transform matrix libraries and OCCT.


bschwind (approved):
LGTM! It would be nice to have a function which sets the entire matrix in one shot, but looking through the occt source it seems like that will complicate things a bit, and there's probably not much of a performance difference, if any.

PR #185, opened 2024-11-17
fix: allow cross compilation under builtin flag v2

fidoriel:
Supersedes #163 while closing #160

### DIFF
--- a/crates/occt-sys/src/lib.rs
+++ b/crates/occt-sys/src/lib.rs
@@ -1,9 +1,46 @@
-use std::path::Path;
+use std::{
+    env::var,
+    path::{Path, PathBuf},
+};
+
+const LIB_DIR: &str = "lib";
+const INCLUDE_DIR: &str = "include";
 
 /// Get the path to the OCCT library installation directory to be
 /// used in build scripts.
 ///
 /// Only valid during build (`cargo clean` removes these files).
-pub fn occt_path() -> &'static Path {
-    Path::new(env!("OCCT_PATH"))
+pub fn occt_path() -> PathBuf {
+    // moves the output into target/TARGET/OCCT
+    // this way its less likely to be rebuilt without a cargo clean
+    Path::new(&var("OUT_DIR").expect("missing OUT_DIR")).join("../../../../OCCT")
+}
+
+/// Build the OCCT library.
+pub fn build_occt() {
+    cmake::Config::new(Path::new(env!("OCCT_SRC_DIR")))
+        .define("BUILD_PATCH", Path::new(env!("OCCT_PATCH_DIR")))
+        .define("BUILD_LIBRARY_TYPE", "Static")
+        .define("BUILD_MODULE_ApplicationFramework", "FALSE")
+        .define("BUILD_MODULE_Draw", "FALSE")
+        .define("USE_D3D", "FALSE")
+        .define("USE_DRACO", "FALSE")
+        .define("USE_EIGEN", "FALSE")
+        .define("USE_FFMPEG", "FALSE")
+        .define("USE_FREEIMAGE", "FALSE")
+        .define("USE_FREETYPE", "FALSE")
+        .define("USE_GLES2", "FALSE")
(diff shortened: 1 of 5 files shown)


crates/opencascade-sys/Cargo.toml
```diff
@@ -13,7 +13,7 @@ cxx = "1"
 [build-dependencies]
 cmake = "0.1"
 cxx-build = "1"
-occt-sys = { version = "0.5.1", optional = true }
+occt-sys = { path = "../occt-sys", optional = true }
```
fidoriel, line 16: @bschwind not sure about this one here

bschwind: Ah, I think we intentionally don't use the local dependency in order for our CI setup to cache it properly. See [this comment](https://github.com/bschwind/opencascade-rs/pull/91#issuecomment-1641572098).

Though we should revisit if that assumption still holds.

fidoriel: so I use the 0.5.1 version?

bschwind: I think we'll have to do it in two parts:

* I'll review the occt-sys changes and once we agree on those we can publish `0.5.2` or `0.6` depending on if it's a breaking change or not
* We can then make the (somewhat trivial) `opencascade-sys` change separately

Sorry for the trouble on that, it's mostly trying to optimize for caching and faster CI build times. I'm also working on getting a windows dev machine set up to more easily be able to test this.

fidoriel: Seems reasonable.

fidoriel: done. Then there will be a follow-up once if released.

bschwind:
Hi @fidoriel, sorry for the delay. I haven't yet been able to get a windows machine up and running and I'll be traveling fairly soon so it might take me some time.

At the very least I'll test this on MacOS and Linux first and we can merge it if that goes well, and then I'll check the windows stuff later.

bschwind (approved):
LGTM! I'll publish this as occt-sys 0.6

bschwind:
@fidoriel I published your changes to occt-sys under [version 0.6](https://crates.io/crates/occt-sys)

crates/occt-sys/src/lib.rs
```diff
@@ -1,9 +1,46 @@
...
-pub fn occt_path() -> &'static Path {
-    Path::new(env!("OCCT_PATH"))
+pub fn occt_path() -> PathBuf {
+    // moves the output into target/TARGET/OCCT
+    // this way its less likely to be rebuilt without a cargo clean
+    Path::new(&var("OUT_DIR").expect("missing OUT_DIR")).join("../../../../OCCT")
+}
+
+/// Build the OCCT library.
+pub fn build_occt() {
```
strohel, line 20: Nobody call this (as of this PR), is that intended?

strohel (commented):
Alright, this makes some sense, the OCCT is built for the "target", not to run as part of the build system. At the same time, I see that other -sys crates in the ecosystem usually perform the cmake build in `build.rs`.

I'm looking at https://github.com/bschwind/opencascade-rs/issues/160 and I wonder (and this may be a silly question): wouldn't it be the solution that `occt-sys` is a _runtime_ dependency of `opencascade-sys`?

For context @fidoriel: I briefly worked on these crates with @bschwind in the past, and he asked me to take a quick look.

crates/occt-sys/src/lib.rs
bschwind: I believe there's a part 2 coming for this PR. We first wanted to merge this to publish the next version of occt-sys since it is not included in the workspace.

fidoriel:
@bschwind was debugging the hell out of my setup but broke every single time. We did not update the crate to 0.6 :)

fidoriel:
Using the branch from https://github.com/bschwind/opencascade-rs/pull/190 I can cross compile and my project runs fine. If there are unknowns, I dont know them.

PR #188, opened 2025-01-07
feature: expose BRepFont builder

mkovaxx:


mkovaxx:
Compiling `opencascade-rs` itself works fine.

But when I use it in an example, I get the following linker errors:
```
Undefined symbols for architecture arm64:
  "StdPrs_BRepFont::RenderGlyph(char32_t const&)", referenced from:
      StdPrs_BRepFont_RenderGlyph(StdPrs_BRepFont&, unsigned int) in libopencascade_sys-91437712f31bd19c.rlib[4635](6fc440bb027e5261-lib.rs.o)
  "StdPrs_BRepFont::StdPrs_BRepFont(NCollection_UtfString<char> const&, Font_FontAspect, double, Font_StrictLevel)", referenced from:
      StdPrs_BRepFont_ctor_from_name(rust::cxxbridge1::String const&, int, double) in libopencascade_sys-91437712f31bd19c.rlib[4635](6fc440bb027e5261-lib.rs.o)
ld: symbol(s) not found for architecture arm64
```

mkovaxx:
Got the linker errors fixed. :)

Now it just fails to find the font:
```
Font_FontMgr, error: unable to find any font!
```

mkovaxx:
My hunch is that the FreeType lib isn't available when building `occt-sys`, so the internal CMake var `HAVE_FREETYPE` ends up being `false`. This causes relevant functions to have empty bodies, e.g.:
https://github.com/Open-Cascade-SAS/OCCT/blob/3d6c21124ad052d1efa7fd6cb80793b24da172dd/src/Font/Font_FTLibrary.cxx#L32-L37

mkovaxx:
As a workaround here, we could read the FreeType info directly and construct each glyph as a `Face` bounded by the outline curves.

The first thing needed to make that happen is the ability to construct Bezier curve `Edge`s:
https://github.com/bschwind/opencascade-rs/pull/191

mkovaxx:
Closing this in favor of https://github.com/bschwind/opencascade-rs/pull/191.

Once constructing Bezier curve `Edge`s is exposed, code can easily be built on top to realize an outline glyph as a B-rep.

PR #190, opened 2025-01-12
update occt sys to 0.6

fidoriel:
follow up to https://github.com/bschwind/opencascade-rs/pull/185 updating occt sys reference to the patched 0.6 version, as discussed in #185


bschwind (approved):
LGTM! I currently only tested on MacOS but it built and ran!

(unrelated to this change) Since the crate size limit was raised for this crate and I re-included some modules in occt-sys, it now takes quite a bit longer to build everything. Oh well. I usually link to the system-installed library anyway, but it's good to be able to build everything mostly self-contained for cross-compilation, and for building a standalone binary.

PR #191, opened 2025-01-14
Construct edge as a Bezier curve from control points

mkovaxx:
The implementation is straightforward; most tricky pieces are already in the codebase.

Added an example of constructing an extruded letter A from bezier curves.

https://github.com/user-attachments/assets/a07c6dca-5d6f-4cf9-a30d-adb1045fc615


examples/src/letter_a.rs
```diff
@@ -0,0 +1,59 @@
...
+    outer.subtract(&inner).into_shape()
+}
+
+fn contour_to_face(contour: Contour) -> Face {
+    let edges: Vec<Edge> = contour
+        .into_iter()
+        .map(|segment_points| {
+            let points = segment_points
+                .into_iter()
+                .map(|(x, y)| DVec3::new(x as f64 / 16.0 - 16.0, y as f64 / 16.0 - 16.0, 0.0));
```
bschwind, line 18: Can you explain the `/ 16.0 - 16.0` ? Is this some sort of font-specific thing?

bschwind (approved):
Looking good, I just have one question.

I hope to expand text functionality a lot more in the future so we can support non Roman characters, complex font shaping, and perhaps also having text follow a path or smoothly along a surface for fancy text embossing.

examples/src/letter_a.rs
mkovaxx: Oh yeah, those are totally "magic numbers"... The contour data used in the example is dumped straight from the TTF file. The `16` is a constant I determined via Eyeball Mark I (TM) to scale the resulting shape to a reasonable size in the viewer, and also more-or-less center it.

mkovaxx: Could bake those constants into the raw outline data if you prefer that.

bschwind: Yeah either baking it or at least pulling them out into named constants would be good.

mkovaxx: Done! :)

bschwind (approved):
LGTM!

PR #195, opened 2025-01-16
More boolean ops for Face and CompoundFace

mkovaxx:
`Face` has `union` and `subtract`. Adding `intersect`, so that e.g. symmetric difference may be computed. Also adding boolean ops for `CompoundFace`.

### DIFF
--- a/crates/opencascade/src/primitives/face.rs
+++ b/crates/opencascade/src/primitives/face.rs
@@ -310,6 +310,20 @@ impl Face {
         CompoundFace::from_compound(compound)
     }
 
+    #[must_use]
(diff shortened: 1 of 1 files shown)


bschwind (approved):
LGTM! This is a draft, is there anything else you wanted to add?

mkovaxx:
> LGTM! This is a draft, is there anything else you wanted to add?

Yep, I think it'd be good to expose the same CSG methods (`union`, `intersect`, `subtract`) on `CompoundFace` itself.

What do you think?

bschwind:
Ohhh yep, if it's as straightforward as this one was then let's add it too.

bschwind (approved):
LGTM!

Issue #176, opened 2024-06-08
Split the cxx bridge file into smaller modules


bschwind:
Here's a good example that shows how this can be done:

https://github.com/KDAB/cxx-qt/pull/350/files

PR #199, opened 2025-03-15
Split C++ bridge into multiple files

### LINKED ISSUE #17: Proposal: Consistent naming / patterns

bschwind:
Closes #17 and #176

TODO

- [x] Use [Self](https://cxx.rs/attributes.html#self) for static methods
- [x] Rename from `ctor` to `new`
- [x] Organize and make naming consistent
- [x] Use C++ templates where applicable to reduce header boilerplate
- [x] Add docs on how to create new bindings


bschwind:
The naming in the bindings isn't _perfect_, but I'm at least happy with the better organization here.

PR #201, opened 2025-03-16
WIP: [kicad-parser] write a KicadBoard

mkovaxx:


mkovaxx:
@bschwind The current data model seems to assume that the `Footprint`X and `Graphic`X structures are identical.
That seems not to be the case:
https://dev-docs.kicad.org/en/file-formats/sexpr-intro/index.html#_footprint_graphics_items
https://dev-docs.kicad.org/en/file-formats/sexpr-intro/index.html#_graphic_items

bschwind:
> The current data model seems to assume that the FootprintX and GraphicX structures are identical.
> That seems not to be the case:

Yep, let's keep these as separate types.

crates/kicad-parser/src/writer.rs
```diff
@@ -0,0 +1,146 @@
+use sexp::{atom_f, atom_s, Sexp};
+
+use crate::{
+    board::{BoardLayer, Footprint, KicadBoard},
+    graphics::{GraphicArc, GraphicCircle, GraphicLine, GraphicRect},
+};
+
+pub fn write_board<W: std::io::Write>(
+    writer: &mut W,
+    board: &KicadBoard,
```
bschwind, line 10: Should this instead be `KicadPcb`? I think my original plan was for `KicadBoard` to hold the actual board data like copper, footprints, silkscreens, etc, while `KicadPcb` would hold a `KicadBoard`, as well as extra data like the version and other metadata.

crates/kicad-parser/src/writer.rs
```diff
@@ -0,0 +1,146 @@
...
+        cons_s("end", point_to_sexp(arc.end_point)),
+        cons_s("layer", layer_to_sexp(&arc.layer)),
+    ];
+    Sexp::List(items)
+}
+
+fn circle_to_sexp(circle: &GraphicCircle) -> Sexp {
+    let items = vec![
+        atom_s("gr_circle"),
+        cons_s("start", point_to_sexp(circle.center_point)),
```
bschwind, line 66:
```suggestion
        cons_s("center", point_to_sexp(circle.center_point)),
```

crates/kicad-parser/src/writer.rs
```diff
@@ -0,0 +1,146 @@
...
+fn point_to_sexp(point: (f64, f64)) -> Sexp {
+    Sexp::List(vec![atom_f(point.0), atom_f(point.1)])
+}
+
+fn layer_to_sexp(layer: &BoardLayer) -> Sexp {
+    let layer_str: &str = layer.into();
+    atom_s(layer_str)
+}
+
+fn position_to_sexp(location: (f64, f64), rotation_degrees: f64) -> Sexp {
```
bschwind, line 133: I might call this `orientation_to_sexp` or `transform_to_sexp`, as I expected `position` to just mean an XY coordinate.

bschwind (commented):
Looking pretty good so far! A few questions:

* What's the scope for this PR, or when do we consider it "done"? When we can write out a fully-functioning `*.kicad_pcb` file?
* It would make sense to have a `write_to_file<P: AsRef<Path>>(file: P)` convenience function
* Pretty printing of the S-expressions would be nice, maybe the `sexp` crate doesn't support that though
* Just for fun I tried exporting a `kicad_pcb` file by reading the `sample.kicad_pcb` file and immediately serializing it again. When trying to read it into KiCAD, I got:

Screenshot 2025-03-20 at 15 38 38

PR #207, opened 2025-05-06
Add section API

tymokvo:
## Summary
- Add wrapper around `BRepAlgoAPI_Section` class
- Add casting function to `wrapper.hxx`
- Add tests
- Add `section` example

## Description
Hello,

I have been evaluating CAD libraries in the rust ecosystem for a project at work. In order to test out some of the capabilities of `opencascade-rs`, I took a stab at implementing a simple wrapper around the `BRepAlgoAPI_Section` class.

The most complicated aspect was dealing with the inherited methods from the `BRepAlgoAPI` class hierarchy. I followed the advice from an [issue on `cxx`](https://github.com/dtolnay/cxx/issues/797#issuecomment-809951817) on how to deal with this, hence the addition to `wrapper.hxx`.

Would this be a useful addition? I know that it's a bit rough right now, I haven't used `cxx` before! Would love to hear any feedback on this direction, though.


bschwind:
Hi @tymokvo, thanks for the PR! And sorry for the late response on it.

I skimmed through the PR and it's looking good, thanks for adding an example as well!

Just FYI, the next thing I hope to accomplish in this repo is to have a more systematic approach to generating the bindings.

This includes logically organizing the OpenCascade modules into equivalently named Rust modules. You can see a manual example of that process in this PR:

https://github.com/bschwind/opencascade-rs/pull/199

But obviously manually creating those are a pain, so I'm also working on generating them by parsing the C++ headers with tree-sitter and writing out the Rust code automatically. That work is happening in the [tree-sitter](https://github.com/bschwind/opencascade-rs/compare/main...tree-sitter) branch.

I just wanted to give you an idea of what I hope this crate will turn into in the future. I'll try to give this a closer review soon.

tymokvo:
The work on the `tree-sitter` branch is looking cool, happy to see that! Writing the bindings manually in one giant file is definitely not super fun.

crates/opencascade/src/section.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct Section {
+    pub(crate) inner: UniquePtr<ffi::BRepAlgoAPI_Section>,
+}
+impl Section {
+    /// Create a new `Section` to intersect `target` by `tool`.
+    pub fn new(target: &Shape, tool: &Shape) -> Section {
+        Section {
+            inner: ffi::BRepAlgoAPI_Section_ctor(
+                target.inner.as_ref().expect("Target was null"),
+                tool.inner.as_ref().expect("Tool was null"),
```
bschwind, line 15: I think this should work:

```suggestion
                &target.inner,
                &tool.inner,
```

crates/opencascade/src/section.rs
```diff
@@ -0,0 +1,76 @@
...
+            ),
+        }
+    }
+
+    /// Get the edges of the resulting intersection.
+    pub fn section_edges(self) -> Vec<Shape> {
+        // TODO: Given that the OCCT name is "SectionEdges", can we return a Vec<Edge>?
```
bschwind, line 22: Looking at the implementation code, it seems likely that would could cast these to `Edge`. But they went with a list of `Shape` instead so I'm not 100% sure it will always return `Edge`s.

crates/opencascade/src/section.rs
```diff
@@ -0,0 +1,76 @@
...
+        let s = Section::new(&a.into_shape(), &b.into_shape());
+
+        let edges = s.section_edges();
+        assert_eq!(edges.len(), 1);
+
+        let itm = edges.first();
+
+        assert!(itm.is_some());
+
+        let s = itm.unwrap();
```
bschwind, line 66: You've already asserted `edges.len() == 1` so this can be simplified:

```suggestion
        let s = edges.first().unwrap();
```

crates/opencascade/src/section.rs
```diff
@@ -0,0 +1,76 @@
...
+
+        assert_eq!(s.shape_type(), ShapeType::Edge);
+
+        for e in s.edges() {
+            // There should be only one edge
+            assert_eq!(e.start_point(), dvec3(0., -0.5, 0.));
+            assert_eq!(e.end_point(), dvec3(0., 0.5, 0.));
+        }
```
bschwind, line 74:
```suggestion
        let e = s.edges().next().expect("There should be only one edge");

        assert_eq!(e.start_point(), dvec3(0., -0.5, 0.));
        assert_eq!(e.end_point(), dvec3(0., 0.5, 0.));
```

This also reminds me I should probably add type conversions between `Shape` and all the other primitive types, maybe a `TryFrom` implementation or just `fn as_edge(&self) -> Option<Edge>`

examples/src/section.rs
bschwind, line 1: Cool example!

bschwind (commented):
Overall this looks great! I left a few comments, and I have one tiny preference to have literal floating point numbers have leading and trailing zeroes:

```rust
0. -> 0.0
.1 -> 0.1
etc.
```

crates/opencascade/src/section.rs
tymokvo: Agreed, I think it seems better to have the caller try that cast rather than create the possibility of a panic from an invalid cast.

examples/src/section.rs
tymokvo: Thank you!

bschwind (approved):
LGTM! I'll merge this once CI passes. Thanks for the PR and thanks for waiting :)

PR #210, opened 2025-06-09
Add gp_Dir.Transform to opencascade-sys

gmorenz:
I need this for transforming normals when exporting meshes.

There's analogous methods here on other types, but I thought it best for now to add what I need. In the future should I add related methods simultaneously?


bschwind:
Hi @gmorenz - this is fine as-is! I'm still deciding on how I want to automate the bindings so this manual process of adding methods will eventually be unnecessary. So I'd say it's fine to add what you need and as long as things build I'll probably accept the PRs :)

PR #212, opened 2025-06-16
Add bounding box support

tymokvo:
## Summary
- Add `bounding_box` module wrapping `Bnd_Box` and `BRepBndLib`
- Add tests
- Add `bounding_box` example

## Description
I need to compute bounding boxes of OCC geometry for a project that uses `opencascade`. Thankfully, OCC has the `BRepBndLib` package that supports this already. This adds a wrapper for the `BRepBndLib::Add` method and the underlying `Bnd_Box` class on which it depends.

To support this and show how to use it, there are a few simple tests and an example that uses the viewer to show the bounded and bounding geometry.

One thing that I didn't understand, but that I assume is important to the underlying algorithms in OCC, is that the `Bnd_Box` has a small `Gap` value that is added to every dimension of the box. I decided to expose this as `gap_vec` to make it easier to work with in the rust side and tried to document it clearly.


bschwind (approved):
LGTM, thanks! I'd like to format the decimal literals in `bounding_box.rs` but I'll handle that.

tymokvo:
Ah thanks! Sorry I missed those!

PR #213, opened 2025-08-21
Add `Shape::transform`

### LINKED ISSUE #194: fn Shape::transform() & co.
Could be mutable
- `transform(&mut self, &DMat4)`
- `translate(&mut self, &DVec3)`
- `rotate(&mut self, &DQuat)`

Or immutable
- `transformed(&self, &DMat4) -> Self`
- `translated(&self, &DVec3) -> Self`
- `rotated(&self, &DQuat) -> Self`

tymokvo:
## Summary
- Expose `gp_Trsf::SetValues`
- Add `transform` module, tests
- Add `gp_trsf(&DMat4) -> gp_Trsf`
- Add `Shape::transform`

## Description
I needed a way to define a `Shape` once and then duplicate that geometry throughout a model space using transform matrices, similar to "blocks" from the CAD world or "instanced rendering" from the GPU programming world.

This PR exposes OCC's implementation of this capability through the `BRepBuilderAPI_Transform` class. This requires using a `gp_Trsf` instance as the transform matrix. There didn't seem to be a convenient way to construct these from a `glam` type, so I added the `transform` module and a simple test to ensure the construction is correct.

One caveat here (that is also noted in the comments) is that the `gp_Trsf` constructor:
- Only allows setting the upper 3x4 matrix 
  - I.e. cannot set the `w` component of any column
- Orthogonalizes the matrix

So, the values of the output can be different than the input based on the orthogonality of the input.

### DIFF
--- a/crates/opencascade/src/primitives/shape.rs
+++ b/crates/opencascade/src/primitives/shape.rs
@@ -7,7 +7,7 @@ use crate::{
     Error,
 };
 use cxx::UniquePtr;
-use glam::{dvec2, dvec3, DVec3};
+use glam::{dvec2, dvec3, DMat4, DVec3};
 use opencascade_sys::ffi;
 use std::path::Path;
 
@@ -631,6 +631,10 @@ impl Shape {
         self.inner.pin_mut().set_global_translation(&location, false);
     }
 
+    pub fn transform(&self, transform: &DMat4) -> Self {
+        Self::from_shape(&self.inner)
+    }
+
     pub fn mesh(&self) -> Result<Mesh, Error> {
         self.mesh_with_tolerance(0.01)
     }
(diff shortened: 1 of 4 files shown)


crates/opencascade/src/primitives/shape.rs
```diff
@@ -631,6 +631,14 @@ impl Shape {
         self.inner.pin_mut().set_global_translation(&location, false);
     }
 
+    /// Copy `self` using the upper 3x4 matrix of `transform`.
+    pub fn transform(&self, transform: &DMat4) -> Self {
+        let t = crate::transform::gp_trsf(transform);
+        // NOTE: Setting the copy argument does not seem to have an effect.
+        let mut res = ffi::BRepBuilderAPI_Transform_ctor(&self.inner, &t, true);
```
bschwind: Could you pull `true` into a named local variable?
```suggestion
        let copy = true;
        let mut res = ffi::BRepBuilderAPI_Transform_ctor(&self.inner, &t, copy);
```

bschwind (approved):
LGTM, thanks for this and sorry for the super long wait!

crates/opencascade/src/primitives/shape.rs
tymokvo: Will do! Sorry for taking so long. Very busy at work the past several weeks!

PR #215, opened 2025-10-16
feat: `Shape::empty()`

### LINKED ISSUE #214: feature: Shape::empty()
## Summary

An empty `Shape` would help simplify many constructions. For example, when a `Shape` collection is combined with `union`, either in a `for` loop, or via a higher-order function like [`Iterator::fold()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold).

## Bikeshed

Alternative names I can think of:
- `Shape::nothing()`
- `Shape::null()`
- `Shape::new()`
- `Shape::void()`

## Practical Example

With `Shape::empty()` we'd be able to write the following:
```rust
let mut composite = Shape::empty();

for part in shape_iterator {
    composite = composite.union(part).into_shape();
}
```

Or for the functionally inclined among us, like this:
```rust
let composite = shape_iterator.fold(Shape::empty(), |acc, item| acc.union(item).into_shape());
```

Currently we have to write this:
```rust
let mut composite: Option<Shape> = None;

for part in shape_iterator {
    if let Some(shape) = &mut composite {
        *shape = shape.union(part).into_shape();
    } else {
        composite = Some(part);
    }
}
```

## Theoretical Background

In general, a [neutral element](https://en.wikipedia.org/wiki/Identity_element) for an associative binary operation is a big win for ease of use. The numbers 0 and 1 make it easier to define sums and products, the empty string helps work with concatenation, a value of `infinity` simplifies the definition of minimum, etc. These are all examples of an algebraic structure called a [monoid](https://en.wikipedia.org/wiki/Monoid), which is closely tied to the notion of [`fold`ing](https://en.wikipedia.org/wiki/Monoid#Monoids_in_computer_science) a sequence or some other "traversable" container such as a tree.

In our case, the associative binary operation is `union`, and `Shape::empty()` would be its neutral element.

## Implementation

It's already possible, along these lines:
```rust
fn empty_shape() -> Shape {
    let shape = Shape::cube(1.0);
    shape.subtract(&shape).into_shape()
}
```

But it would be nice to instead expose the constructor [`TopoDS_Shape()`](https://dev.opencascade.org/doc/refman/html/class_topo_d_s___shape.html#a52fca99449feb70f1f83eb8bb028274f) which "Creates a NULL Shape referring to nothing."

I'd be happy to take this on. :)

mkovaxx:


mkovaxx:
To test, run:
```
cargo run --release --bin viewer -- --example pentafoil
```

Screenshot 2025-10-16 at 16 26 16

bschwind (approved):
LGTM, thanks for this!

Though wow, that example takes quite awhile to build the model.

mkovaxx:
> LGTM, thanks for this!
> 
> Though wow, that example takes quite awhile to build the model.

It's a good stress test for sure. ;)

PR #219, opened 2026-02-14
Add BSpline curve data extraction from Edge

janbraiins:
## Summary

- Add `Edge::bspline_data()` method to extract degree, control points (poles), flat knot vector, and weights from B-spline curves
- Includes automatic conversion of non-BSpline curves via `GeomConvert`
- Uses bulk extraction functions for efficiency (single FFI call for all poles, knots, or weights instead of per-element calls)


janbraiins:
If the bspline_data() is too much, I can just leave the pure [FFI extension](https://github.com/bschwind/opencascade-rs/pull/219/commits/59207a537e094e33452c26002db16b3e4545dfb9)

PR #221, opened 2026-02-23
LLM-written binding generator covering most of Open CASCADE - discussion?

sethml:
So, um, I wanted to vibe-code a STL to STEP converter, and decided Rust was the language to use, and Open CASCADE was the library to use, and found your project, then discovered that the bindings weren't very complete, and well, things got a bit out of hand.

I've now vibe-coded a binding generator that creates Rust bindings for most of OCCT. I also had it port crates/opencascade and your examples to the new bindings, to demonstrate that they actually work. I'm going to start work on my STL to STEP converter next.

I'm honestly not sure what to do with my fork of your project now. I discovered partway through that I was basically re-creating bindgen, but specialized for OCCT. I think maybe this makes sense, since it covers the conventions and quirks of OCCT. Or maybe it's just a big pile of AI slop. I do think it's probably useful for people who want to use the raw OCCT library from Rust.

I'd love your feedback on what to do with it. I can imagine a few possibilities:
1. You love it, and pull into your repo, and update the crates. I'd be fine with that.
2. You hate it and ignore it. I'd be fine with that too.
3. I push it under some new name on crates.io.
4. I keep it as a personal thing.

It does need some work to be shareable, I think. It generates different code depending on platform and occt version. Maybe I need to move away from checking in the generate files and instead have them be created during the build process based on your particular OCCT library install. And it could probably use somewhat better documentation and probably more testing - I suspect that some APIs get incorrect bindings.

I'd love to hear your thoughts.


bschwind:
Well this is certainly interesting, and I'll run the CI just for kicks, but I can't really feasible review or accept this into the repo unfortunately.

I have thought about using an LLM to handle the bindings in some way, because yeah it is pretty mundane drudgery for a programming task. I had looked into a lean tree-sitter based approach but didn't make a whole lot of progress on that, so in my next attempt I was going to have another look at `autocxx`. Whatever I end up going with though, I'd like to have a handle on how it all works in case I need to debug or maintain it, and for hobby projects I'm unfortunately not interested in the "just let LLMs deal with it" approach, sorry.

That being said, I recognize my project probably shows up quite often when you look for opencascade-related things in Rust, and I know the bindings are quite incomplete. For now I would recommend option 3 or 4 on your list. The code in this repo is licensed the same as opencascade itself, so feel free to push your own thing on crates.io or whatever you'd like to do.

Regarding the task you actually want to do, making an STL to STEP converter, I'm curious: what program is giving you STL files and not STEP? OpenSCAD?

sethml:
Thanks for taking a prompt look. The CI tests probably fail because the bindings are generated on MacOS, and a few of the types differ on Linux. Generating the wrappers only takes a few seconds, at least on a fast system, so I should probably just make them be auto-generated as part of the build process. I assume that's what bindgen does, although I've never used it.

And I totally understand your reaction to this - I'd probably have the same. I think the path I'll take is to rename it on github to avoid confusion with your project, use it for my project, try to clean up the code generation, perhaps try to modularize the bindings (it's pulling in a lot of rarely-used stuff now), maybe add some more testing, and then push a new crate to crates.io.

Why do I want to convert STL to STEP? Mostly as an interesting test challenge for AI coding, to help me understand how it works. And because I'm sick of people publishing interesting/useful models on Printables and such as STL without STEP files or original CAD. A converter that makes assumptions that the STL was generated by a CAD program, thus all vertices should lie almost exactly on a corresponding analytical surface, ought to be able to produce a nice STEP representation, which is then easy to edit to do things like resize holes or change dimensions.

I'm retired from a background of writing very careful and low-level C++, and I'm pretty impressed with what I can accomplish with AI coding so far, with just a few minutes of attention every few hours to guide it. I was pretty skeptical at first, but I have to admit that it really changes the picture for a lot of software development. I do worry about the next generation of developers - I think you need to have a pretty good idea of how software works and good development practices to avoid creating a total mess with AI, and I'm not sure how people will develop that knowledge and skill if they can just prompt an LLM rather than learning. It's a brave new world.

PR #222, opened 2026-04-15
WIP: TDF/OCAF bindings PoC

Ben-PH:
Adds Rust bindings for OCCT's TDF and TNaming subsystems, surfacing stable topological shape identity across model operations. This allows for applications to be developed in accordance with OCCT expectations, and avoiding preventable topo-naming nightmares.

Work is incomplete, and the proposed API would benefit from the careful attention of an experienced eye in review.

## What's in

- `TDF_Data`, `TDF_Label`, `TDF_Transaction`, `TDF_Delta` with lifetime-enforced ownership chain — `'doc` lifetime prevents labels and transactions outliving their document, enforced at compile time
- `TDF_Data::Undo` with consume-by-value delta semantics — applying a delta is a one-way operation; the inverse (redo) delta is returned
- `TNaming_Builder` bindings — `Generated` (with and without predecessor), `Modify`, `Delete`, `Select` — exposed through a safe Rust API taking `&Shape` rather than raw `&TopoDS_Shape`
- `TnamingBuilder` borrows the open `TdfTransaction` for its full lifetime — the borrow checker prevents committing while a builder is live
- `TnamingNamedShape::get()` returning the current active shape, and `original_shape()` returning the recorded old shape — both coherent across undo/redo boundaries
- `Shape::translation() -> DVec3` for location querying without leaking FFI types
- Full undo/redo round-trip test asserting active shape state at each stage via `get()`
- Compile-fail doc-tests demonstrating lifetime constraints

## Known gaps flagged in comments

- `TdfTransaction::open` return value needs newtyping (nested transaction index)
- Builder construction on an unopened transaction cannot be enforced by the borrow checker — future closure-based API will make this impossible to misorder
- `build.rs` link additions for TDF/TNaming libraries may be needed depending on build configuration

Disclaimer: Much of this work was done with a lot of heavy lifting coming from LLM assistance, but without LLM directly touching the codebase.

### DIFF
--- a/crates/opencascade-sys/include/wrapper.hxx
+++ b/crates/opencascade-sys/include/wrapper.hxx
@@ -83,6 +83,9 @@
 #include <TDF_Label.hxx>
 #include <TDF_Delta.hxx>
 #include <TDF_Transaction.hxx>
+#include <TNaming_Builder.hxx>
+#include <TNaming_NamedShape.hxx>
+#include <TNaming_Tool.hxx>
 #include <gp.hxx>
 #include <gp_Ax2.hxx>
 #include <gp_Ax3.hxx>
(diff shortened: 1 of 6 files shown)


Ben-PH:
Before having solic confidence in these changes, I'll make use of the in my cad application. That should be completed within a day or two

PR #224, opened 2026-06-04
Add translate/rotate/scale/mirror transform operations to Shape

torusJKL:
`set_global_translation` mutates `&mut self` in place and only supports translation via a low-level Location mechanism.
 The new methods (translated, rotated, scaled, mirrored):

- Are non-mutating: &self → Self, returning a new copy (the original is untouched)
- Use BRepBuilderAPI_Transform with copy=true, which is the standard, OpenCASCADE API for shape transforms
- Work generically on any Shape, not narrowed to Wire

### DIFF
--- a/crates/opencascade/src/primitives/shape.rs
+++ b/crates/opencascade/src/primitives/shape.rs
@@ -9,7 +9,7 @@ use crate::{
 use cxx::UniquePtr;
 use glam::{dvec2, dvec3, DVec3};
 use opencascade_sys as ffi;
-use std::path::Path;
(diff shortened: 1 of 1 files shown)


bschwind (approved):
Thanks for the PR! The code makes sense, I think it just needs a `cargo +nightly fmt` run.

torusJKL:
@bschwind I pushed an update that should now pass the formatting check.

bschwind:
Thanks again!

torusJKL:
Thanks for the quick merge!

PR #225, opened 2026-06-05
Add safe downcasts from Shape to Wire/Face/Solid

torusJKL:
Shape already has upcasts via From<Wire/Face/Solid> for Shape, but no safe way to go the other direction.

The macro assumes every subshape type follows:
- `TopoDS::$TYPE`, OCCT's static downcast function (e.g. TopoDS::Wire, TopoDS::Face)
- `ShapeType::$TYPE`, the type discriminator in the Rust enum (e.g. ShapeType::Wire, ShapeType::Face)
- `from_<snake_case_type>`, the pub(crate) constructor (e.g. Wire::from_wire, Face::from_face)
- `as_<snake_case_type>` / `expect_<snake_case_type>`, the generated downcast methods

Out of 7 subshapes Vertex is the odd one out, it only has `Vertex::new(point)`, so the macro would fail with a "method not found" error.

Optionally a `pub(crate) fn from_vertex` would make Vertex compatible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2199,6 +2199,7 @@ dependencies = [
  "glam 0.24.2",
  "kicad-parser",
  "opencascade-sys",
+ "paste",
  "thiserror 1.0.64",
 ]
 
--- a/crates/opencascade/Cargo.toml
+++ b/crates/opencascade/Cargo.toml
@@ -12,6 +12,7 @@ cxx = "1"
 opencascade-sys = { version = "0.2", path = "../opencascade-sys" }
 glam = { version = "0.24", features = ["bytemuck"] }
 kicad-parser = { path = "../kicad-parser" }
+paste = "1"
 thiserror = "1"
 
 [features]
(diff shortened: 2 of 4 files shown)


bschwind (commented):
This makes sense! Would it be alright if we add these without a macro though? I think it would be justified if there were dozens of types to convert between, but we just have a few. It will be slight repetition but I think it's more straightforward to read.

torusJKL:
@bschwind since this is only for 3 types the code duplication is indeed not much.
Maybe a macro could become interesting should there be more types in the future.

Please feel free to squash the commits if you like the direct implementations without the macro more.

bschwind:
Yeah it's close to the threshold of abstracting it to a macro, but the code as it is now is more straightforward to read and easier for tools to find (in my opinion, at least).

I squash all commits from PRs into `main` so that will happen either way. Thanks again!

bschwind (approved):

PR #226, opened 2026-06-07
write multiple shapes to one STEP file

torusJKL:
OCCT can write multiple shapes as individual items to a single step file.
The current write_step method only accepts a single shape.

This change includes:
- Add write_all_step to transfers multiple shapes to a single step writer
- Refactor write_step to delegate to write_all_step
- Add StepWriteNoShapes error for empty input (OCCT does not write empty step files and silently fails)
- Add StepWriteTransferFailed error to distinguish transfer failures


bschwind (approved):
LGTM, thanks!

torusJKL:
Thanks for merging!

PR #227, opened 2026-06-24
Expose BRepBuilderAPI_MakeFace::Add for faces with holes

torusJKL:
Enabling creation of faces with cutouts (regular cut boolean does not work with coplanar faces).
Takes an outer wire and a slice of hole wires and crates a new face.


bschwind (approved):
LGTM, nice!

PR #228, opened 2026-06-29
Shape inspection

torusJKL:
This PR adds inspection/query capabilities across shape types.
I included all the different types in one PR since they are all about inspection of properties but I can split them based on shape type if that is preferable.

The following can be inspected:

#### Solid
- volume(), center_of_mass(), surface_area()

#### Face / Surface
- SurfaceType enum with face type detection
- Surface geometry: periodicity, continuity, degrees, knots, poles, evaluation (d0, d1, d2)

#### Edge
- length(), tangent_at(), arc_center(), radius()

#### Wire
- length(), is_closed()

#### Shape
- center_of_mass()

### DIFF
--- a/crates/opencascade-sys/include/b_rep.hxx
+++ b/crates/opencascade-sys/include/b_rep.hxx
@@ -32,6 +32,4 @@ inline std::unique_ptr<Handle_Poly_Triangulation> BRep_Tool_Triangulation(const 
       new opencascade::handle<Poly_Triangulation>(BRep_Tool::Triangulation(face, location)));
 }
 
-inline bool BRep_Tool_IsClosed(const TopoDS_Wire &wire) {
-  return BRep_Tool::IsClosed(wire);
-}
+inline bool BRep_Tool_IsClosed(const TopoDS_Wire &wire) { return BRep_Tool::IsClosed(wire); }
--- a/crates/opencascade-sys/include/b_rep.hxx
+++ b/crates/opencascade-sys/include/b_rep.hxx
@@ -7,6 +7,7 @@
 #include <TopoDS_Edge.hxx>
 #include <TopoDS_Face.hxx>
 #include <TopoDS_Vertex.hxx>
+#include <TopoDS_Wire.hxx>
 #include <bindings_common.hxx>
 #include <gp_Pnt.hxx>
 
@@ -29,4 +30,8 @@ inline std::unique_ptr<Handle_Poly_Triangulation> BRep_Tool_Triangulation(const 
                                                                           TopLoc_Location &location) {
   return std::unique_ptr<Handle_Poly_Triangulation>(
       new opencascade::handle<Poly_Triangulation>(BRep_Tool::Triangulation(face, location)));
+}
+
+inline bool BRep_Tool_IsClosed(const TopoDS_Wire &wire) {
+  return BRep_Tool::IsClosed(wire);
 }
--- a/crates/opencascade-sys/include/b_rep_adaptor.hxx
+++ b/crates/opencascade-sys/include/b_rep_adaptor.hxx
@@ -1,7 +1,17 @@
 #include <BRepAdaptor_Curve.hxx>
 #include <bindings_common.hxx>
+#include <gp_Circ.hxx>
 #include <gp_Pnt.hxx>
+#include <gp_Vec.hxx>
 
 inline std::unique_ptr<gp_Pnt> BRepAdaptor_Curve_value(const BRepAdaptor_Curve &curve, const Standard_Real U) {
   return std::unique_ptr<gp_Pnt>(new gp_Pnt(curve.Value(U)));
+}
+
+inline void BRepAdaptor_Curve_D1(const BRepAdaptor_Curve &curve, const Standard_Real U, gp_Pnt &P, gp_Vec &V1) {
+  curve.D1(U, P, V1);
+}
+
+inline std::unique_ptr<gp_Circ> BRepAdaptor_Curve_Circle(const BRepAdaptor_Curve &curve) {
+  return std::unique_ptr<gp_Circ>(new gp_Circ(curve.Circle()));
 }
--- a/crates/opencascade-sys/include/geom.hxx
+++ b/crates/opencascade-sys/include/geom.hxx
@@ -2,10 +2,14 @@
 #include <Geom_BezierCurve.hxx>
 #include <Geom_BezierSurface.hxx>
 #include <Geom_Circle.hxx>
+#include <Geom_ConicalSurface.hxx>
 #include <Geom_CylindricalSurface.hxx>
 #include <Geom_Plane.hxx>
+#include <Geom_SphericalSurface.hxx>
 #include <Geom_Surface.hxx>
+#include <Geom_ToroidalSurface.hxx>
 #include <Geom_TrimmedCurve.hxx>
+#include <gp_Ax3.hxx>
 #include <gp_Circ.hxx>
 #include <bindings_common.hxx>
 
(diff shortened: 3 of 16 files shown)


bschwind:
@torusJKL Thanks for another PR, I'll try to get to reviewing this soon, it'll take a bit of time as it's a larger PR.

crates/opencascade-sys/include/b_rep.hxx
```diff
@@ -30,3 +31,5 @@ inline std::unique_ptr<Handle_Poly_Triangulation> BRep_Tool_Triangulation(const
   return std::unique_ptr<Handle_Poly_Triangulation>(
       new opencascade::handle<Poly_Triangulation>(BRep_Tool::Triangulation(face, location)));
 }
+
+inline bool BRep_Tool_IsClosed(const TopoDS_Wire &wire) { return BRep_Tool::IsClosed(wire); }
```
bschwind, line 35: It's slightly clumsier in this case (because you have to cast a Wire to a Shape), but you can bind to static class functions like this:

```rust
#[Self = "BRep_Tool"]
pub fn IsClosed(shape: &TopoDS_Shape) -> bool;
```

Just noting that as an option though, this code is fine as-is.

crates/opencascade/tests/edge_inspection.rs
```diff
@@ -0,0 +1,72 @@
...
+    let length = edge.length();
+    approx_equal(length, std::f64::consts::PI * 10.0);
+}
+
+#[test]
+fn line_tangent_at_midpoint() {
+    let edge = Edge::segment(DVec3::new(0.0, 0.0, 0.0), DVec3::new(10.0, 0.0, 0.0));
+    let tangent = edge.tangent_at(0.5, PositionMode::Parameter);
+    let expected = DVec3::new(1.0, 0.0, 0.0);
+    assert!(tangent.distance_squared(expected) < 1e-6, "expected {expected:?}, got {tangent:?}");
```
bschwind, line 23: A dot product being approximately equal to `1.0` would probably be a more appropriate check here.

crates/opencascade/tests/wire_inspection.rs
```diff
@@ -0,0 +1,28 @@
...
+    let edge = Edge::segment(DVec3::ZERO, DVec3::new(10.0, 0.0, 0.0));
+    let wire = Wire::from_edges([&edge]);
+    assert!(!wire.is_closed());
+}
+
+fn approx_equal(a: f64, b: f64) {
+    let diff = (a - b).abs();
+    let rel = diff / b.abs().max(1e-12);
+    assert!(rel < 1e-4 || diff < 1e-6, "expected {b}, got {a} (diff={diff}, rel={rel})");
+}
```
bschwind, line 28: Let's move this into a shared test module that `edge_inspection` can also use.

crates/opencascade-sys/include/b_rep_adaptor.hxx
```diff
@@ -1,7 +1,17 @@
...
 #include <gp_Pnt.hxx>
+#include <gp_Vec.hxx>
 
 inline std::unique_ptr<gp_Pnt> BRepAdaptor_Curve_value(const BRepAdaptor_Curve &curve, const Standard_Real U) {
   return std::unique_ptr<gp_Pnt>(new gp_Pnt(curve.Value(U)));
 }
+
+inline void BRepAdaptor_Curve_D1(const BRepAdaptor_Curve &curve, const Standard_Real U, gp_Pnt &P, gp_Vec &V1) {
+  curve.D1(U, P, V1);
+}
```
bschwind, line 13: This one can fortunately be auto-bound

```suggestion

```

In `b_rep_adaptor.rs`, you can write:

```rust
#[cxx_name = "D1"]
pub fn d1(self: &BRepAdaptor_Curve, u: f64, point: Pin<&mut gp_Pnt>, vec: Pin<&mut gp_Vec>);
```

And then in `edge.rs`, it can be called like so:

```rust
curve.d1(param, point.pin_mut(), vec.pin_mut());
```

In general I try to avoid as much C++ code as possible in the wrappers. For things like wrapping code in `unique_ptr`, or handles, it's hard to avoid, but for simpler types you can usually get cxx-rs to automatically bind things.

crates/opencascade-sys/src/b_rep_adaptor.rs
```diff
@@ -15,6 +17,13 @@ mod inner {
         pub fn FirstParameter(self: &BRepAdaptor_Curve) -> f64;
         pub fn LastParameter(self: &BRepAdaptor_Curve) -> f64;
         pub fn BRepAdaptor_Curve_value(curve: &BRepAdaptor_Curve, u: f64) -> UniquePtr<gp_Pnt>;
+        pub fn BRepAdaptor_Curve_D1(
+            curve: &BRepAdaptor_Curve,
+            u: f64,
+            point: Pin<&mut gp_Pnt>,
+            vec: Pin<&mut gp_Vec>,
+        );
```
bschwind, line 25: As noted in another comment, this can become:

```rust
#[cxx_name = "D1"]
pub fn d1(self: &BRepAdaptor_Curve, u: f64, point: Pin<&mut gp_Pnt>, vec: Pin<&mut gp_Vec>);
```

crates/opencascade/src/primitives.rs
```diff
@@ -267,3 +273,36 @@ impl From<JoinType> for ffi::geom_abs::GeomAbs_JoinType {
...
+            SurfaceType::Torus => write!(f, "Torus"),
+            SurfaceType::BezierSurface => write!(f, "BezierSurface"),
+            SurfaceType::BSplineSurface => write!(f, "BSplineSurface"),
+            SurfaceType::SurfaceOfRevolution => write!(f, "SurfaceOfRevolution"),
+            SurfaceType::SurfaceOfExtrusion => write!(f, "SurfaceOfExtrusion"),
+            SurfaceType::OffsetSurface => write!(f, "OffsetSurface"),
+            SurfaceType::OtherSurface => write!(f, "OtherSurface"),
+        }
+    }
+}
```
bschwind, line 308:
```suggestion

```

We can drop this entire `impl`. It's not used, as far as I can tell, and it's just re-implementing what `#[derive(Debug)]` already does.

crates/opencascade/src/primitives/surface.rs
```diff
@@ -31,4 +31,131 @@ impl Surface {
 
         Self { inner }
     }
+
+    pub fn surface_type(&self) -> SurfaceType {
+        let dynamic_type = ffi::geom::DynamicType(&self.inner);
+        let name = ffi::standard::type_name(dynamic_type);
+
+        match name.as_str() {
```
bschwind, line 39: Instead of matching on strings, let's wrap `self.inner` in a `GeomAdaptor_Surface`, and call `GetType()` on it, so we can work with the `GeomAbs_SurfaceType` C++ enum instead.

We can add a binding to the C++ enum via `geom_abs.rs` (there's already an example of doing the same thing in that file with `GeomAbs_CurveType` and `GeomAbs_JoinType`).

crates/opencascade/src/primitives/surface.rs
```diff
@@ -31,4 +31,131 @@ impl Surface {
...
+            "Geom_BezierSurface" => SurfaceType::BezierSurface,
+            "Geom_BSplineSurface" => SurfaceType::BSplineSurface,
+            "Geom_SurfaceOfRevolution" => SurfaceType::SurfaceOfRevolution,
+            "Geom_SurfaceOfExtrusion" => SurfaceType::SurfaceOfExtrusion,
+            "Geom_OffsetSurface" => SurfaceType::OffsetSurface,
+            _ => SurfaceType::OtherSurface,
+        }
+    }
+
+    pub fn axis(&self) -> Option<(DVec3, DVec3)> {
```
bschwind, line 54: I'd prefer a struct with named fields instead of a tuple with two vectors in them. We could maybe put it in `primitives.rs` for now, or define some new `geometry.rs` module.

crates/opencascade/src/primitives/surface.rs
```diff
@@ -31,4 +31,131 @@ impl Surface {
...
+        let surface_type = self.surface_type();
+        match surface_type {
+            SurfaceType::Cylinder => {
+                let cyl =
+                    ffi::geom::new_HandleGeomCylindricalSurface_from_HandleGeomSurface(&self.inner);
+                if cyl.IsNull() {
+                    return None;
+                }
```
bschwind, line 62: This PR maybe isn't the right place, but I'm thinking an enum that holds data might be nicer for these functions. Something like:

```rust
pub enum SurfaceType {
    Cylinder(Handle_Geom_CylindricalSurface),
    Sphere(Handle_Geom_SphericalSurface),
   ...
}
```

And then we could convert a `Surface` to its specific enum variant one time (and do the null check), and then be able to call things like `axis()` or `radius()` without having to do further checks. But I'll play around with that idea later.

crates/opencascade/src/primitives/surface.rs
```diff
@@ -31,4 +31,131 @@ impl Surface {
...
+                Some(ffi::geom::HandleGeomConicalSurface_RefRadius(&cone))
+            },
+            SurfaceType::Torus => {
+                let torus =
+                    ffi::geom::new_HandleGeomToroidalSurface_from_HandleGeomSurface(&self.inner);
+                if torus.IsNull() {
+                    return None;
+                }
+                Some(ffi::geom::HandleGeomToroidalSurface_MajorRadius(&torus))
+            },
```
bschwind, line 143: It might be worth noting we return the "major" radius in the case of a torus, since it has two radii to consider.

This does make me think that having more specialized types makes sense, so that if you have a given concrete type like `Torus`, you know you can call things like `major_radius()` and `minor_radius()`, without needing to have them wrapped in an `Option`.

crates/opencascade/src/primitives/face.rs
```diff
@@ -394,6 +395,42 @@ impl Face {
...
+        let surface = Surface { inner: handle };
+        surface.radius()
+    }
+
+    pub fn surface_type(&self) -> SurfaceType {
+        let surface = ffi::b_rep::BRep_Tool_Surface(&self.inner);
+        let dynamic_type = ffi::geom::DynamicType(&surface);
+        let name = ffi::standard::type_name(dynamic_type);
+
+        match name.as_str() {
```
bschwind, line 415: Same here, let's use the C++ enum instead of the dynamic type name.

crates/opencascade/src/primitives/face.rs
```diff
@@ -570,6 +608,56 @@ mod tests {
...
+        let shape = Shape::cylinder_radius_height(3.0, 10.0);
+        let has_cylinder = shape.faces().any(|f| f.surface_type() == SurfaceType::Cylinder);
+        assert!(has_cylinder, "Expected at least one cylindrical face");
+    }
+
+    #[test]
+    fn test_surface_type_sphere() {
+        let shape = Shape::sphere(5.0).build();
+        let has_sphere = shape.faces().any(|f| f.surface_type() == SurfaceType::Sphere);
+        assert!(has_sphere, "Expected at least one spherical face");
```
bschwind, line 628: A sphere should have exactly one spherical face, right?

crates/opencascade/src/primitives/edge.rs
```diff
@@ -174,6 +174,55 @@ impl Edge {
...
+    pub fn tangent_at(&self, position: f64, mode: PositionMode) -> DVec3 {
+        let curve = ffi::b_rep_adaptor::BRepAdaptor_Curve_new(&self.inner);
+        let param = match mode {
+            PositionMode::Parameter => position,
+            PositionMode::Length => {
+                let first = curve.FirstParameter();
+                let last = curve.LastParameter();
+                first + position * (last - first)
+            },
+        };
```
bschwind, line 198: Do we really want `PositionMode` at all here? The API is confusing, and I don't see any tests demonstrating what it's useful for. It interpolates between the first and last parameter of the curve, but when would you use that?

crates/opencascade-sys/include/geom.hxx
```diff
@@ -43,3 +49,76 @@ inline std::unique_ptr<gp_Pnt> HandleGeomCurve_Value(const Handle_Geom_Curve &cu
...
+new_HandleGeomSphericalSurface_from_HandleGeomSurface(const Handle_Geom_Surface &surface) {
+  Handle_Geom_SphericalSurface handle = opencascade::handle<Geom_SphericalSurface>::DownCast(surface);
+  return std::unique_ptr<Handle_Geom_SphericalSurface>(new opencascade::handle<Geom_SphericalSurface>(handle));
+}
+
+inline double HandleGeomSphericalSurface_Radius(const Handle_Geom_SphericalSurface &sphere) { return sphere->Radius(); }
+
+inline const gp_Ax3 &HandleGeomSphericalSurface_Position(const Handle_Geom_SphericalSurface &sphere) {
+  return sphere->Position();
+}
```
bschwind, line 88: Can these be auto-bound?

crates/opencascade-sys/include/geom.hxx
```diff
@@ -43,3 +49,76 @@ inline std::unique_ptr<gp_Pnt> HandleGeomCurve_Value(const Handle_Geom_Curve &cu
...
+  return std::unique_ptr<Handle_Geom_ConicalSurface>(new opencascade::handle<Geom_ConicalSurface>(handle));
+}
+
+inline double HandleGeomConicalSurface_RefRadius(const Handle_Geom_ConicalSurface &cone) { return cone->RefRadius(); }
+
+inline double HandleGeomConicalSurface_SemiAngle(const Handle_Geom_ConicalSurface &cone) { return cone->SemiAngle(); }
+
+inline const gp_Ax3 &HandleGeomConicalSurface_Position(const Handle_Geom_ConicalSurface &cone) {
+  return cone->Position();
+}
```
bschwind, line 104: Same here, let's try to auto-bind these.

crates/opencascade-sys/include/geom.hxx
```diff
@@ -43,3 +49,76 @@ inline std::unique_ptr<gp_Pnt> HandleGeomCurve_Value(const Handle_Geom_Curve &cu
...
+  return torus->MajorRadius();
+}
+
+inline double HandleGeomToroidalSurface_MinorRadius(const Handle_Geom_ToroidalSurface &torus) {
+  return torus->MinorRadius();
+}
+
+inline const gp_Ax3 &HandleGeomToroidalSurface_Position(const Handle_Geom_ToroidalSurface &torus) {
+  return torus->Position();
+}
```
bschwind, line 124: And here

crates/opencascade-sys/src/geom.rs
```diff
@@ -74,6 +87,41 @@ mod inner {
             geom_surface_handle: &Handle_Geom_Surface,
         ) -> UniquePtr<Handle_Geom_Plane>;
 
+        pub fn new_HandleGeomCylindricalSurface_from_HandleGeomSurface(
+            geom_surface_handle: &Handle_Geom_Surface,
+        ) -> UniquePtr<Handle_Geom_CylindricalSurface>;
+        pub fn HandleGeomCylindricalSurface_Radius(cyl: &Handle_Geom_CylindricalSurface) -> f64;
+        pub fn HandleGeomCylindricalSurface_Position(
+            cyl: &Handle_Geom_CylindricalSurface,
+        ) -> &gp_Ax3;
```
bschwind, line 96: Sorry the code for `Handles` is pretty atrocious right now, I need to find a better way to abstract this, maybe a macro or something.

bschwind (commented):
Thanks for being patient! In general the changes are good but there are some improvements we can make to reduce the C++ boilerplate, and remove things like runtime string matching.

